PR #3, opened 2017-06-02
Add tools for security files generation

mikaelarguedas:
addresses https://github.com/ros2/ros2/pull/340#discussion_r119451953

connects to ros2/ros2#339

PR #4, opened 2017-06-14
Sros2 dirk

dirk-thomas:
This PR targets the branch of #3. It updates the script to use `ros2cli`. In the last commit I cleanup the style as well as some Python constructs.

@mikaelarguedas It would be great if you could test if the verbs are still doing what they are supposed to do.


sros2/api/__init__.py
```diff
@@ -123,63 +121,64 @@ def create_keystore(args):
     print(args)
 
     if not os.path.exists(root):
-        print("creating directory: %s" % root)
-        os.makedirs(root)
+        print('creating directory: %s' % root)
+        os.makedirs(root, exist_ok=True)
```
dirk-thomas: FYI: `exist_ok=True`

sros2/api/__init__.py
```diff
@@ -337,16 +334,15 @@ def create_key(args):
...
-        print("bad character in requested key name: %s" % name)
+        print('bad character in requested key name: %s' % name)
         return False
-    print("creating key for node name: %s" % name)
+    print('creating key for node name: %s' % name)
 
     key_dir = os.path.join(root, name)
-    if not os.path.exists(key_dir):
-        os.makedirs(key_dir)
+    os.makedirs(key_dir, exist_ok=True)
```
dirk-thomas: FYI: `exist_ok=True`

sros2/api/__init__.py
```diff
@@ -123,63 +121,64 @@ def create_keystore(args):
...
     else:
-        print("found signed governance file, not creating a new one!")
+        print('found signed governance file, not creating a new one!')
 
     # create index file
     index_path = os.path.join(root, 'index.txt')
     if not os.path.isfile(index_path):
-        open(index_path, 'a').close()
+        with open(index_path, 'a'):
+            pass
```
dirk-thomas: FYI

sros2/api/__init__.py
```diff
@@ -402,81 +398,14 @@ def create_key(args):
 
 
 def list_keys(args):
-    for root, dirs, files in os.walk(args.ROOT):
-        if root == args.ROOT:
-            for d in dirs:
-                print("%s" % d)
+    for name in os.listdir(args.ROOT):
```
dirk-thomas: FYI `listdir` vs. `walk`.

mikaelarguedas:
Thanks! I'll give it a shot soon

sros2/api/__init__.py
codebot: Ah, thanks. This is good to know. Much cleaner.

sros2/verb/create_key.py
```diff
@@ -0,0 +1,34 @@
...
+from ros2pkg.verb import VerbExtension
+from sros2.api import create_key
+
+
+class CreateKeyVerb(VerbExtension):
+    """Create key."""
+
+    def add_arguments(self, parser, cli_name):
+        arg = parser.add_argument('ROOT', help='root path of keystore')
+        arg.completer = DirectoriesCompleter()
```
mikaelarguedas: actually even if I have it installed it crashes. DirectoriesCompleter has been added in v0.8.5 and xenial ships with 0.8.1 (yeah I know zesty ships with 1.7 and pip with 1.8...) so I think the check on the import rule should be stricter

dirk-thomas: In the case of an `ImportError` I set the class name to `None`. That obviously can't be called. Can you please try it with the following code instead:

```
DirectoriesCompleter = lambda: None
```

mikaelarguedas: Yep the lambda trick works. I did split the imports to be able to use FilesCompleter even if DirectoryCompleter is not found

dirk-thomas: :+1:

dirk-thomas:
> Is it expected that I don't have bash completion neither for the verbs nor the arguments of the verbs?

I assume you have `argcomplete` installed. After sourcing the setup file of the workspace (`.bash` of course) you should have completion for "everything": commands, verbs, argument (e.g. package names, topic names), as well as options (e.g. `--help`).

mikaelarguedas:
> >  Is it expected that I don't have bash completion neither for the verbs nor the arguments of the verbs?

> I assume you have argcomplete installed. After sourcing the setup file of the workspace (.bash of course) you should have completion for "everything": commands, verbs, argument (e.g. package names, topic names), as well as options (e.g. --help).

Actually its working now that the imports are split. Not sure what I missed yesterday. Thanks!

PR #12, opened 2017-07-03
Split docs and update content

mikaelarguedas:


mikaelarguedas:
As usual feel free to push typos and english language fixes directly

SROS2_Windows.md
```diff
@@ -0,0 +1,85 @@
...
+
+
+## Run the demo
+
+Open a new terminal:
+
+
+```
+cd C:\dev\ros2
+call setup.bat
```
dirk-thomas: How is the `setup.bat` file getting into `C:\dev\ros2`?

SROS2_Linux.md
```diff
@@ -0,0 +1,201 @@
...
+
+```bash
+ros2 security create_key demo_keys talker
+ros2 security create_key demo_keys listener
+```
+
+### Define the SROS2 environment variables
+
+```bash
+export ROS_SECURITY_ROOT_DIRECTORY=$(pwd)/demo_keys
```
dirk-thomas: I would suggest to use `~/sros2_demo` instead of `${pwd}` since the user might just change the cwd accidentally.

mikaelarguedas:
> Quite some content seems to be duplicated between Linux and OS X. I am not sure that makes sense?

I went with the unified approach to start with. Given that installation steps as well as the supported features diverge, there would be a lot of "(Linux only)" instructions making it hard to read for the MacOS users so I decided to split them

PR #13, opened 2017-07-05
add libssl-dev as an exec dependency

mikaelarguedas:
already in rosdep database here: https://github.com/ros/rosdistro/blob/master/rosdep/base.yaml#L2715

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -9,6 +9,8 @@
 
   <depend>ros2cli</depend>
 
+  <exec_depend>libssl-dev</exec_depend>
+
   <test_depend>ament_copyright</test_depend>
   <test_depend>ament_flake8</test_depend>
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
oh good point, I'll try it to make sure that works and update it + put that back in review when confirmed

dirk-thomas:
Is this targeting the beta 2 release (which would require retagging / re-packaging)?

mikaelarguedas:
nope this has been packages with a dependency on `libssl-dev` I just forgot to update the manifest accordingly here

mikaelarguedas:
Ok this works fine with `openssl`. Only fastrtps will need to be compiled with `libssl-dev`. Thanks @tfoote for pointing it out

mikaelarguedas:
ready for :eyes:

PR #14, opened 2017-07-05
update dockerfile to use beta2 binaries

mikaelarguedas:


mikaelarguedas:
ready for review

mikaelarguedas:
Is there a will to keep a Dockerfile building the code from source as an example ? I know that several users mentioned wanting to have a dockerfile for ROS2 development / testing including @chapulina yesterday.
This doesnt prevent from merging this as is it's more of a follow-up discussion

dirk-thomas:
Currently we don't maintain any Dockerfiles for ROS 2. Instead we provide from-source instructions only. To minimize the effort to maintain both and keeping them in sync I would favor only supporting one. That can be the from source instructions (so users have to copy-n-paste the lines into a Dockerfile) or we could provide a Dockerfile which contains plenty of comments (which then can serve as the from-source instructions).

mikaelarguedas:
>  That can be the from source instructions (so users have to copy-n-paste the lines into a Dockerfile) or we could provide a Dockerfile which contains plenty of comments (which then can serve as the from-source instructions).

:+1: I like the idea and don't feel strongly about one rather than the other

dirk-thomas:
The second option even has the benefit that we could automate the build which will ensure that our instructions work as intended.

mikaelarguedas:
> The second option even has the benefit that we could automate the build which will ensure that our instructions work as intended.

Agreed.
I was just wondering if the first one would not be more readable / conventional for users (not everybody is familiar with docker and could be surprised to have to remove all these `RUN` statements when just being willing to copy paste from the instructions like on the other platforms), that's why I didnt vouch for one or the other

PR #16, opened 2017-07-17
update xml to match spec + connext 53

mikaelarguedas:
Should be merged after Connext 5.3 is released + https://github.com/ros2/rmw_connext/pull/238 and https://github.com/ros2/rmw_connext/pull/237 are merged

### DIFF
--- a/sros2/api/__init__.py
+++ b/sros2/api/__init__.py
@@ -134,8 +134,12 @@ def create_governance_file(path, domain_id):
             <domains>
               <id>%s</id>
             </domains>
-            <!-- <allow_unauthenticated_join>false</allow_unauthenticated_join> SPECIFIED IN DDS SECURITY XSD -->
-            <allow_unauthenticated_participants>false</allow_unauthenticated_participants> <!-- SPECIFIED IN DDS SECURITY BUT NOT IN THE OFFICIAL XSD, IMPLEMENTED BY RTI DDSSEC-130 and RTI XSD http://community.rti.com/schema/5.3.0/dds_security_governance.xsd -->
+            <!-- SPECIFIED IN DDS SECURITY XSD -->
+            <!-- <allow_unauthenticated_join>false</allow_unauthenticated_join> -->
+            <!-- SPECIFIED IN DDS SECURITY BUT NOT IN THE OFFICIAL XSD,
+            IMPLEMENTED BY RTI DDSSEC-130 and RTI XSD:
+            http://community.rti.com/schema/5.3.0/dds_security_governance.xsd -->
+            <allow_unauthenticated_participants>false</allow_unauthenticated_participants> 
             <enable_join_access_control>true</enable_join_access_control>
             <discovery_protection_kind>ENCRYPT</discovery_protection_kind>
             <liveliness_protection_kind>ENCRYPT</liveliness_protection_kind>
--- a/sros2/api/__init__.py
+++ b/sros2/api/__init__.py
@@ -139,7 +139,7 @@ def create_governance_file(path, domain_id):
             <!-- SPECIFIED IN DDS SECURITY BUT NOT IN THE OFFICIAL XSD,
             IMPLEMENTED BY RTI DDSSEC-130 and RTI XSD:
             http://community.rti.com/schema/5.3.0/dds_security_governance.xsd -->
-            <allow_unauthenticated_participants>false</allow_unauthenticated_participants> 
+            <allow_unauthenticated_participants>false</allow_unauthenticated_participants>
             <enable_join_access_control>true</enable_join_access_control>
             <discovery_protection_kind>ENCRYPT</discovery_protection_kind>
             <liveliness_protection_kind>ENCRYPT</liveliness_protection_kind>
--- a/sros2/api/__init__.py
+++ b/sros2/api/__init__.py
@@ -128,11 +128,14 @@ def create_governance_file(path, domain_id):
         f.write("""\
 <?xml version="1.0" encoding="UTF-8"?>
 <dds xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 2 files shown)


sros2/api/__init__.py
```diff
@@ -128,21 +128,24 @@ def create_governance_file(path, domain_id):
...
     <domain_access_rules>
         <domain_rule>
-            <domain_id>%s</domain_id>
-            <allow_unauthenticated_join>FALSE</allow_unauthenticated_join>
-            <enable_join_access_control>TRUE</enable_join_access_control>
+            <domains>
+              <id>%s</id>
+            </domains>
+            <!-- <allow_unauthenticated_join>false</allow_unauthenticated_join> SPECIFIED IN DDS SECURITY XSD -->
+            <allow_unauthenticated_participants>false</allow_unauthenticated_participants> <!-- SPECIFIED IN DDS SECURITY BUT NOT IN THE OFFICIAL XSD, IMPLEMENTED BY RTI DDSSEC-130 and RTI XSD http://community.rti.com/schema/5.3.0/dds_security_governance.xsd -->
```
wjwwood: This line is reallllly long, maybe consider putting the comment on the line above, and splitting it after the `:` and before the url.

mikaelarguedas: good point, fixed in https://github.com/ros2/sros2/pull/16/commits/e1e2c0bf1471888e8ff136242e3fec033c48ba6e

mikaelarguedas:
Note: this PR also fixes the access control that have been broken since we merged the namespace storage in partitions. This still assumes empty namespace (`rt` partitions). If we stick with the current namespace approach we will need to provide a Python API to convert a fully qualified ROS name into a partition and a topic name string to genertate these xml files. @wjwwood @Karsten1987 FYI

wjwwood:
That function would need to be provided by the rmw implementation or better yet it this access control file would be generated by the rmw implementation, not by this tool. Because if the security file is not the same for Fast-RTPS and Connext, for example, then you don't want this tool to be encapsulating those differences.

mikaelarguedas:
> [...] you don't want this tool to be encapsulating those differences.

Agreed. Ideally the resulting files would be the same for all rmw_implementation (following DDS-Security spec) but specific to the protocol this way if we have an rmw_implementation not relying on DDS, a specific set of files would be generated for that transport.
It's an analog problem to xml files for QoS for example. Ideally the same XML file would work for all DDS implementations.

It's definitely not a blocker for this PR I just wanted to bring it up

mikaelarguedas:
this is waiting for feedback from RTI and can be merged once 5.3.0 is available on their website

dirk-thomas:
Does this patch still work with Connext 5.2 or only with 5.3?

mikaelarguedas:
This patch will work only with connext 5.3 given that connext changed an xml tag for access control (that matches the official spec and not the OMG examples)
https://github.com/ros2/sros2/pull/16/files#diff-164a61093789ed771b5f88b17b5e4468R142

Then the current version of the generated xml files doesn't work since we merged the namespacing. I can split this into 2 PR to first fix the current behavior with 5.2.3 and delay this one for connext 5.3

dirk-thomas:
While Connext 5.3 is available for customers it is not "generally" available yet. The launch is schedule for Sep. 12th. Once that is the case and users as well as our CI system can fetch Connext 5.3 we can consider upgrading to the new version. Since this will be after ROS 2 Beta 3 we won't spend any effort on making the binary packages work with Connext out-of-the-box before then.

sros2/api/__init__.py
```diff
@@ -306,32 +321,50 @@ def create_permission_file(path, name, domain_id, permissions_dict):
...
+        </subscribe>
 """
 
     permission_str += """\
-    </allow_rule>
-    <default>DENY</default>
-  </grant>
-</permissions>
+      </allow_rule>
+      <default>DENY</default>
```
ruffsl: Never mind what I said @mikaelarguedas , from testing this branch, your default deny rule works just fine.

mikaelarguedas: cool :+1:

mikaelarguedas:
based on https://github.com/ros2/sros2/pull/16#issuecomment-324696878 I'm putting this back in backlog until connext 5.3 is available for download

mikaelarguedas:
Ok I rebased and updated this. Tested locally using the new connext 5.3.0 from the RTI website and it works in both cases:
- wildcards
- specific access control rules

Note that RTI provides binaries with security on OSX and Windows we will also need to validate these platforms

Some changes will need to be made to test_communication to be able to test this on the farm:
- update tests to catch connext exceptions on access control violations (diffferent from fastrtps that just doesnt match the participants and hangs)
- regenerate key/certs/permission files to integrate the xml changes of this PR
- add new tests to test access control policies

Not sure if we want to do this incrementally and merge this before iterating on the rest or hold this until all the work is done and merge everything at once.

ruffsl:
> regenerate key/certs/permission files to integrate the xml changes of this PR

I've been working on refactoring the keystore and metafile generation, so you might not want spend effort in duplicating that.

mikaelarguedas:
> I've been working on refactoring the keystore and metafile generation, so you might not want spend effort in duplicating that.

Here I'm referring to just calling "create_key" and "create_permission" to update the files used in https://github.com/ros2/system_tests/tree/master/test_communication/test/test_security_files, not touching the generation code.

No worries I'm not planning on touching the generation itself and am looking forward to used the revamped / much more awesome one :)

mikaelarguedas:
Good surprise of the day: security tests with connext passed on Windows and MacOS even if we never tried it before!
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2582/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3301/)

mikaelarguedas:
As discussed offline, I'm going to merge this as is and iterate on the tests from https://github.com/ros2/system_tests/pull/230 separately.

PR #17, opened 2017-08-22
remove linter dependency from Dockerfile

mikaelarguedas:
addresses https://github.com/osrf/docker_images/pull/68#discussion_r134572819

### DIFF
--- a/sros2_docker/docker/sros2/Dockerfile
+++ b/sros2_docker/docker/sros2/Dockerfile
@@ -25,9 +25,7 @@ RUN apt-get -qq update && \
       ros-r2b2-sros2 \
       vim
 
-RUN pip3 install flake8 \
-      flake8-import-order \
-      argcomplete
+RUN pip3 install argcomplete
 
 # setup ros2 workspace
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Is the expectation that users don't build their own packages using these Docker containers?

mikaelarguedas:
Which container are you referring to ? This one was made for demo purposes and should be discarded/removed once we ship containers for users to use so I think there is no need to install one specific linter while we don't install any of the other ones.
For the "official" ones that we may ship in the future, it's a good question, I think that we should either install all common test dependencies or none, but not only flake8.

ruffsl:
I'd keep the list indented, but that's my dockerfile format preference.

mikaelarguedas:
I don't think that single-line commands should be indented, though I agree that the previous list should have been.
I don't feel strongly about it, feel free to push a fix to that branch if you think that an indented list makes more sense here

ruffsl:
I find it easier to investigate multi-line lists when using the github blame feature to track when a package was added or removed. But it's no big.

PR #20, opened 2017-09-06
Fix connext for node with default partition

mikaelarguedas:
This doesnt take into accounts namespaces yet but fixes access control for nodes without namespace.
Works for connext 5.2.4.

This fix is temporary and will be overriden by https://github.com/ros2/sros2/pull/16 once connext 5.3 is released

### DIFF
--- a/sros2/api/__init__.py
+++ b/sros2/api/__init__.py
@@ -132,17 +132,17 @@ def create_governance_file(path, domain_id):
     <domain_access_rules>
         <domain_rule>
             <domain_id>%s</domain_id>
-            <allow_unauthenticated_join>FALSE</allow_unauthenticated_join>
-            <enable_join_access_control>TRUE</enable_join_access_control>
+            <allow_unauthenticated_join>false</allow_unauthenticated_join>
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
ready for review

mikaelarguedas:
Yeah sorry it's not obvious to evaluate, this is a minimal version of https://github.com/ros2/sros2/pull/16.

Up until now we were ignoring partitions for access control. That was working fine until we implemented namespaces. This PR adds a partition wildcard for [publishers](https://github.com/ros2/sros2/pull/20/files#diff-164a61093789ed771b5f88b17b5e4468R325) and [subscribers](https://github.com/ros2/sros2/pull/20/files#diff-164a61093789ed771b5f88b17b5e4468R329) in the wildcard access control xml and our default partition for [publishers and subscribers](https://github.com/ros2/sros2/pull/20/files#diff-164a61093789ed771b5f88b17b5e4468R313) where we do enforce access control.

In the future we should get the namespacing from the provided yaml file and construct partitions accordingly but as we discussed in the past maybe the rmw implementations should generated those themselvses in case they split the FQN differently so I left it out for now

PR #24, opened 2017-09-12
Add internal topics (without partition) to default allow rule

dhood:
The topic wildcard + partition wildcard doesn't match

This should fix the `RTI_Security_AccessControl_check_create_datareader:endpoint not allowed: no rule found; default DENY` error when using default access policies. @mikaelarguedas to confirm


mikaelarguedas:
Rebased and resolved conflicts. I can confirm that it fixes the wildcards, thanks @dhood for testing the access control tutorial!

PR #25, opened 2017-09-12
install/setup.bat -> local_setup.bat

sloretz:
install/setup.bat does not exist

### DIFF
--- a/SROS2_Windows.md
+++ b/SROS2_Windows.md
@@ -64,7 +64,7 @@ Open a new terminal:
 
 ```
 cd C:\dev\ros2
-call install\setup.bat
+call local_setup.bat
 ros2 run demo_nodes_py talker
 ```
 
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
Yeah that's a tricky one, the presence or not of the install folder depends if you build from source or installed from binaries...

But both `local_setup.bat` and `setup.bat` should be present. I'm surprised it;s not the case for you

sloretz:
`setup.bat` exists, but there's no `install` folder. What's the difference between `setup.bat` and `local_setup.bat` anyways?

mikaelarguedas:
`setup.bat` will source underlay workspaces while `local_setup.bat` sources only this workspace

Not sure what's the best way foarwrd is on this one, looks a bit too much to have to make a different case to specify how to source a workspace if building from source or binaries... But I don;t see a better alternative

sloretz:
I think I misunderstood what `install/setup.bat` meant. How about `<path_to_ros2_install>/setup.bat` ?

mikaelarguedas:
sounds good to me

PR #29, opened 2017-12-07
explicitly call out setting the variables

tfoote:
It's commented above, but this leaves the optimization to leave them out for those who know more rather than just having it not work.


mikaelarguedas:
Sounds good to me. I originally didnt put it as I expected Windows users to set these variables in their environment and thus have them defines for any new console they open after that.

PR #33, opened 2018-02-21
Fix access control for ardent

mikaelarguedas:
connects to ros2/sros2#32

This is the minimal set of changes to get access control working for Connext with the default talker/listener.

This doesn't aim to add feature to support service Access Control. Just ensuring that the nodes can be crerated and use enforced access control for pub sub.

What this does:
- add the clock topic to allowed subscription by default
- add the parameter_events topic to allowed publications and subscriptions by default
- add publication for all default parameter service Request topics
- add subscription for all default parameter service Reply topics

PR #34, opened 2018-02-21
remove what now appears to be obsolete DCPS whitelisting

mikaelarguedas:
WIP: needs testing

### DIFF
--- a/sros2/api/__init__.py
+++ b/sros2/api/__init__.py
@@ -403,13 +403,7 @@ def create_permission_file(path, name, domain_id, permissions_dict):
         </%s>
 """ % (pubsubtag, partition_string, topics_string, pubsubtag)
 
-    # DCPS* is necessary for builtin data readers
     permission_str += """\
-        <subscribe>
-          <topics>
-            <topic>DCPS*</topic>
-          </topics>
-        </subscribe>
       </allow_rule>
       <default>DENY</default>
     </grant>


mikaelarguedas:
This has been reported by @ruffsl (thanks!) in https://github.com/ros2/sros2/pull/33#discussion_r169754779

We originally added it for Connext 5.2.X because we couldn't get simple participants created without it.
Tried it locally with Connext 5.3.0 and I confirm that this is not necessary anymore to get talker / listener working

mikaelarguedas:
this has been rebased and conflicts resolved

dhood:
Since I wouldn't have expected this behaviour to change between connext versions (not that it's impossible), I'm wondering what could have changed on our side to not make these lines required anymore. FWIU it might be a sign that discovery isn't encrypted anymore while it used to be, so it'd be good to investigate to confirm that's not the case.

Are you aware of anything specific that changed? I know there was a new xml spec, maybe the defaults changed for something? If not, it could be because of changes on the ROS side (e.g. partitions) or just an accidental regression.

Do we know the relationship between these lines and the "discovery protection" lines (e.g. https://github.com/ros2/sros2/blob/573b47f540b8580614187ed3a503dd224d64e821/sros2/api/__init__.py#L144)? Perhaps enabling discovery protection on set of topics doesn't work with partitions, and now that we use partitions for namespaces that's why we don't need the `DCPS` lines anymore?

Under the assumption that ensuring discovery protection is important, it seems to me that we should check that the current behaviour still matches what we're expecting. I can confirm that the nodes still connect with this change, but haven't investigated _why_ they're connected without this change.

mikaelarguedas:
Did you look at the discovery packet on wireshark by any chance ? (not asking you to do extra work just making sure we don't look at the same thing)

mikaelarguedas:
Looking at the frames it looks like the discovery data is encrypted. The participant coming up can still be seen with the name of the created participant but any topic related information is encrypted (name, type etc).
There are some mentions to BuiltinTopics fixes related to security in the [Connext 5.3.0 Security release notes](https://community.rti.com/static/documentation/connext-dds/5.3.0/doc/manuals/connext_dds/dds_security/RTI_SecurityPlugins_ReleaseNotes.pdf) (though I didnt look in details). As 5.3.0 is the first official version supported security, I'm inclined to consider the previous 5.2.4 behavior regarding builtin topics to be a bug that has been fixed since

dhood:
This one stands out as a potential reason (maybe adding the DCPS lines was a workaround for it):

> 2.3.2.7 Changes in RTI Endpoint Discovery parameter 0x8018 in Security Plugins 5.2.6 and Higher
Security Plugins 5.2.6 introduced changes in the mapping of the EndpointSecurity attributes to the Endpoint Discovery Parameter 0x8018 with respect to release 5.2.5.
Secure Endpoints will now perform consistency checks for the endpoint security attributes. This will allow for early detection of incompatible secure configurations in the Governance configuration that may have resulted in serialization/deserialization errors and/or no communication.

But whether or not that's the reason, I totally take your point about 5.2.3 not quite being what can be considered stable, so it's no longer a surprise that these lines aren't needed now; if the discovery data is encrypted then this PR LGTM!

Issue #32, opened 2018-02-20
Minimal permission set for a talker and listener pair in the demo_cpp example


ruffsl:
I'm trying to determine the minimal set of permissions for a talker and listener pair in the demo_cpp example in order to evaluate keymint's provisioning of permissions. I've checked the discovery data, and I looks like I've added everything they'd need in the permissions document, but apparently I've missed something the client library is needing, as this inevitably causes the failure in creating the DDS participant.

## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 16.04
- Installation type:
  - [ros2.repos](https://github.com/ros2/ros2/blob/f21563fb6ccbd648fb69bf9673e8b2f1bf429aae/ros2.repos)
- Version or commit hash:
  - ef1bec6bb9a68e4b18c298da651332763054525c
- DDS implementation:
  - RTI Connext
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

I've made a simple example that can quickly be create using keymint_tools. See the example section in the repo README for more details: https://github.com/keymint/keymint_tools#example

You can also find the resulting keystore here: https://github.com/ruffsl/keymint_ws/tree/7783f87783e27efcee80be7f6189ded1bce5869e
For example. here is the permissions file for the talker before signing:
https://github.com/ruffsl/keymint_ws/blob/7783f87783e27efcee80be7f6189ded1bce5869e/build/talker/permissions.xml

#### Expected behavior

From looking at the discovery data, we can collect the number of resources advertised. I'd expect collpacing this back down into the permissions files would be sufficient, but perhaps there are some requested resources from the client library to the dds vender that are not being advertised during a non-secure session, or are not captured by my DDS session monitor under secure sessions without access control.

For good measure, one can simple add this XML snippet to the beginning of the `allow_rule` for the `grant` specific to the node within the keymint package source, then rebuild the keymint package and all will work as expected. However, this use of wildcarding nullifies what we are trying to achieve here.

``` xml

                <publish>
                    <topics>
                        <topic>*</topic>
                    </topics>
                    <partitions>
                        <partition>*</partition>
                    </partitions>
                </publish>
                <subscribe>
                    <topics>
                        <topic>*</topic>
                    </topics>
                    <partitions>
                        <partition>*</partition>
                    </partitions>
                </subscribe>
```

#### Actual behavior

[...]

PR #36, opened 2018-02-28
updates to #33

dhood:
note this is targeted at https://github.com/ros2/sros2/pull/33 not master

@mikaelarguedas I added 416f176e28ec455d5c3da0977e50c5264ef53322 since otherwise the parameter_events lines in the sample policy are being over-written, which might "surprise" users. However, since this change impacts the sample policy, I wanted to check that it's a reasonable change conceptually before pushing to the branch (there might be other side effects somewhere that you're aware of)

### DIFF
--- a/examples/sample_policy.yaml
+++ b/examples/sample_policy.yaml
@@ -3,14 +3,10 @@ nodes:
     topics:
       chatter:
         allow: s # can subscribe to chatter
-      parameter_events:
-        allow: p # can publish on parameter_events
   talker:
     topics:
       chatter:
         allow: p # can publish on chatter
-      parameter_events:
-        allow: p # can publish on parameter_events
   listener_py:
     topics:
       #'*':
--- a/examples/sample_policy.yaml
+++ b/examples/sample_policy.yaml
@@ -3,10 +3,14 @@ nodes:
     topics:
       chatter:
         allow: s # can subscribe to chatter
+      parameter_events:
+        allow: p # can publish on parameter_events
   talker:
     topics:
       chatter:
         allow: p # can publish on chatter
+      parameter_events:
+        allow: p # can publish on parameter_events
   listener_py:
     topics:
       #'*':
--- a/sros2/api/__init__.py
+++ b/sros2/api/__init__.py
@@ -312,6 +312,7 @@ def create_permission_file(path, name, domain_id, permissions_dict):
     topic_dict = permissions_dict['topics']
     if topic_dict:
         # add rules for automatically created ros2 topics
+        # TODO(mikaelarguedas) remove this hardcoded handling for default topics
         # TODO(mikaelarguedas) update dictionary based on existing rule
         # if it already exists (rather than overriding the rule)
         topic_dict['parameter_events'] = {'allow': 'ps'}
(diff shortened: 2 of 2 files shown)


dhood:
If a user is specifying that the node should only be able to subscribe to parameter events then there might be a valid reason for that. Similarly if they only want to be able to publish the clock, they should be allowed to specify that. That the startup crashes if they try to do something else is reasonable behaviour in that case IMO; I don't think that the union is the right thing because users might have a reason for the restriction.

This hypothetical situation doesn't apply for the talker demo, so at the very least the existing specification for the parameter_events topic should be removed from the sample policies, because it is misleading that it is explicitly set as subscribe but actually publish/subscribe permissions are being set. But having thought about it, I don't think union makes sense for general use, now or in the future

dhood:
well, I mean, we can go either way, it's a matter of convention I suppose. I have in mind that the policy yaml file lets you set the final permissions, as opposed to the "additional" permissions, but this isn't necessarily the only "right" approach!

mikaelarguedas:
> If a user is specifying that the node should only be able to subscribe to parameter events then there might be a valid reason for that. 

I think our default behavior should match the default permissions required when you inherit from / instanciate a node. As pointed out in https://github.com/ros2/sros2/issues/32#issue-298470925 the error message returned by Connext is not very explicit so it's hard to know what ROS entity is the culprit, especially if it is an entity created by ROS2 under the hood and not one created in users' code.

I would expect users to only define what they use in their node and the rest "should" work. The nature of the change proposed here is going to break any user having an existing yaml file derived from the examples we gave in previous releases because we won't add our new requirements to the resulting permissions file.
I agree that the current state of #33 doesn't allow users using custom nodes (without the clock, parameter or service interface) to generate permission files with only what they need. But this PR adresses only part of that concern (the parameter related permissions are still added to the permissions) with the potential of breaking users that were using policies derived from our examples. 

> This hypothetical situation doesn't apply for the talker demo, so at the very least the existing specification for the parameter_events topic should be removed from the sample policies, because it is misleading that it is explicitly set as subscribe but actually publish/subscribe permissions are being set.

I totally agree with that. Can you modify this PR to only make the changes to the yaml file?

Again the fact that #33 doesnt handle the case where users declare nodes publishing to clock is unfortunate but I think is acceptable in the scope of the ardent patch release as this topic wasnt used anywhere before.

A bigger refactor of the way this script works, allowing users to also opt-out of the defaultly created ROS interfaces (topics and services) would be a better way to solve it, but I think is out os scope of this minimal fix for ardent.

mikaelarguedas:
Thanks!

dhood:
Updated it so it won't break people that have a policy based off the sample that had only publish permissions for paramter_events.

> As pointed out in #32 (comment) the error message returned by Connext is not very explicit so it's hard to know what ROS entity is the culprit, especially if it is an entity created by ROS2 under the hood and not one created in users' code.


FYI (for you or anyone reading this that gets stuck) what I found useful in debugging during this PR was adding `rcutils_logging_set_logger_level("rcl", RCUTILS_LOG_SEVERITY_DEBUG);` to the main so you can see which topic is trying to be created when you get the error.

PR #40, opened 2018-03-06
add special service rule only if not wildcarding everything

mikaelarguedas:
this restricts the scope of the special rules we add. Mainly, we should not add them if we wildcard everything as these rules will never be evaluated.

### DIFF
--- a/sros2/api/__init__.py
+++ b/sros2/api/__init__.py
@@ -340,6 +340,48 @@ def create_permission_file(path, name, domain_id, permissions_dict):
           </topics>
         </%s>
 """ % (tag, 'rt', topic_name, tag)
+        # TODO(mikaelarguedas) remove this hardcoded handling for default parameter topics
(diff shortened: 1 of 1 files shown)


sros2/api/__init__.py
```diff
@@ -340,6 +340,48 @@ def create_permission_file(path, name, domain_id, permissions_dict):
           </topics>
         </%s>
 """ % (tag, 'rt', topic_name, tag)
+        # TODO(mikaelarguedas) remove this hardcoded handling for default parameter topics
+        # TODO(mikaelarguedas) remove the need for empty partition (required for Connext at startup),
```
dhood: fyi line length fixup in https://github.com/ros2/sros2/commit/0d4fc19c33adfde5ab9f50092bcf6131ea98e13e

mikaelarguedas: thanks!

PR #44, opened 2018-04-23
add missing enable_liveliness_protection tag in governance files

mikaelarguedas:
this tag is defined by the spec and was missing from our generated files.
Connext ignores it but Fast-RTPS parser flags the governance file as invalid if this tag is missing.

A follow up PR should be created on `test_security` to update the security files accordingly.


mikaelarguedas:
@ros2/team this and connected PR are ready for review

PR #45, opened 2018-04-30
remove partitions

mikaelarguedas:
This moves the hack for default created topics and services from partition to topic names to match the code changes from ros2/ros2#476.

Note that this now enforces that all created topics have an empty partition

Connects to ros2/ros2#476

### DIFF
--- a/sros2/api/__init__.py
+++ b/sros2/api/__init__.py
@@ -342,9 +342,6 @@ def create_permission_file(path, name, domain_id, permissions_dict):
         </%s>
 """ % (tag, 'rt/' + topic_name, tag)
         # TODO(mikaelarguedas) remove this hardcoded handling for default parameter topics
-        # TODO(mikaelarguedas) remove the need for empty partition
-        # (required for Connext at startup),
-        # see https://github.com/ros2/sros2/issues/32#issuecomment-367388140
         service_topic_prefixes = {
             'Request': 'rq/%s/' % name,
             'Reply': 'rr/%s/' % name,
--- a/sros2/api/__init__.py
+++ b/sros2/api/__init__.py
@@ -353,13 +353,13 @@ def create_permission_file(path, name, domain_id, permissions_dict):
             'list_parameters',
             'describe_parameters',
         ]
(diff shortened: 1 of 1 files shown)


sros2/api/__init__.py
```diff
@@ -334,20 +334,20 @@ def create_permission_file(path, name, domain_id, permissions_dict):
...
+            <partition></partition>
           </partitions>
           <topics>
             <topic>%s</topic>
           </topics>
         </%s>
-""" % (tag, 'rt', topic_name, tag)
+""" % (tag, 'rt/' + topic_name, tag)
         # TODO(mikaelarguedas) remove this hardcoded handling for default parameter topics
         # TODO(mikaelarguedas) remove the need for empty partition
```
dhood: this todo can be removed, right?

sros2/api/__init__.py
```diff
@@ -356,16 +356,17 @@ def create_permission_file(path, name, domain_id, permissions_dict):
             'list_parameters',
             'describe_parameters',
         ]
-        for key, value in service_topic_prefixes.items():
-            if key == 'Request':
+        for service_direction, topic_prefix in service_topic_prefixes.items():
+            if service_direction == 'Request':
```
mikaelarguedas: Thanks for looking into that. I couldnt find a "good" name when I refactored so I didnt change it.

I don't think that "service_direction" is what we mean here. For example, a service server would subscribe to a `Request` topic.

I made another attempt at renaming the variable in https://github.com/ros2/sros2/pull/45/commits/70bc265690e547a03600538436ff4d236ce740c5 trying to match the "topic_prefix" variable introduced here. Feedback welcome

dhood: yep, good point, that name looks good!

mikaelarguedas:
merging this as all dependeant PRs have been merged.

Access control is currently not working since we introduced the [parameter servers starting automatically](https://github.com/ros2/rclcpp/pull/478) with the node. As this is out of the scope of this PR I will address it in a follow-up

PR #53, opened 2018-06-22
Update docs for bouncy leveraging remapping for demo

dhood:
@mikaelarguedas can you review please and if it looks good we can propagate to the other platform docs.

### DIFF
--- a/SROS2_Linux.md
+++ b/SROS2_Linux.md
@@ -177,7 +177,7 @@ To do this, we will use the sample policy file provided in `examples/sample_poli
 First, we will copy this sample policy file into our keystore:
 
 ```bash
-curl -sk https://raw.githubusercontent.com/ros2/sros2/ardent/examples/sample_policy.yaml -o ./demo_keys/policies.yaml
+curl -sk https://raw.githubusercontent.com/ros2/sros2/master/examples/sample_policy.yaml -o ./demo_keys/policies.yaml
 ```
 
 And now we will use it to generate the XML permission files expected by the middleware:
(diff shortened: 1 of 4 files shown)


dhood:
yep will do!

SROS2_Linux.md
```diff
@@ -108,8 +112,19 @@ In another terminal (after preparing the terminal as previously described), we w
...
+These nodes will be communicating using authentication and encryption!
+If you look at the packet contents on e.g. Wireshark, the messages will be encrypted.
+
 Note: You can switch between the C++ (demo_nodes_cpp) and Python (demo_nodes_py) packages arbitrarily.
 
+These nodes are able to communicate because we have created the appropriate keys and certificates for them.
+However, other nodes will not be able to communicate, e.g. the following invocation will fail to start a node with a name that is not associated with valid keys/certificates:
+
+```bash
+# This will fail because the node name does not have valid keys/certificates
```
ruffsl: Clarification: This would fail given the node would then be requesting system topics for a namespace it wasn't provisioned in the Secure DDS atricats, i.e. that include `not_talker` in the namespace, not necessary because the node's name is directly attributed into the public certificate itself, e.g. its common name or otherwise. Or has this changed?

mikaelarguedas: This will actually fail because the namespace in the keystore doesnt exist so it wouldnt even go to the state of checking permissions but fail at the higher level saying "these files don't exist".

Regarding the name matching the common name, I think the current implementation would work as I believe the matching of the common name is done on the name prior to remapping (aka `talker`) though the permissions (if not wildcard) will not match the topics created after participant initialization as these will use the remapped node name

Issue #41, opened 2018-03-10
ros2 security create_permission should abort when a policy file does not exist


sloretz:
## Bug report

**Required Info:**

- Operating System:
  - Windows 10
- Installation type:
  - binaries
- Version or commit hash:
  - Ardent patch 2
- DDS implementation:
  -  Fast-RTPS
- Client library (if applicable):
  - N/A

#### Steps to reproduce issue

```
md C:\dev\ros2\sros2_demo
cd C:\dev\ros2\sros2_demo
ros2 security create_keystore demo_keys
ros2 security create_key demo_keys talker
ros2 security create_permission demo_keys talker this_policy_file_does_not_exist.yaml
```

#### Expected behavior
ros2 security create_permission should refuse to continue if the policy file does not exist.

#### Actual behavior
The policy file is overwritten with permissions that allow everything. stdout from the command is indistinguishable from a run with a real policy file

PR #58, opened 2018-06-22
Windows tutorial tweaks

sloretz:
* `set` instead of `export`
* use `bat` for cmd.exe highlighting
* Remove some trailing backticks
* Add note about `RANDFILE`
* Remove `-s` option from curl (my installation of curl fails because it doesn't support https, and slient hides this)

### DIFF
--- a/SROS2_Windows.md
+++ b/SROS2_Windows.md
@@ -27,7 +27,7 @@ If you don't have OpenSSL installed, please see follow [these instructions](http
 
 We will now create a folder to store all the files necessary for this demo:
 
-```bash
+```bat
 md C:\dev\ros2\sros2_demo
 ```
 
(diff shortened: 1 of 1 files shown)


SROS2_Windows.md
```diff
@@ -107,13 +116,13 @@ To do this, we will use the sample policy file provided in `examples/sample_poli
 
 First, we will copy this sample policy file into our keystore:
 
-```bash
-curl -sk https://raw.githubusercontent.com/ros2/sros2/ardent/examples/sample_policy.yaml -o .\demo_keys\policies.yaml
+```bat
+curl -k https://raw.githubusercontent.com/ros2/sros2/ardent/examples/sample_policy.yaml -o .\demo_keys\policies.yaml
```
mikaelarguedas: Is this because we want additional printing if the download fail?
Or did something else motivate this change?

sloretz: In my case `-s` was hiding an error saying curl was not built with https support, and I almost didn't notice due to #41. By chance I saw the file wasn't there when deleting the demo_keys directory for another RMW implementation.

Issue #49, opened 2018-06-04
generated xml files don't validate schema


mikaelarguedas:
Based on the current xsd: http://www.omg.org/spec/DDS-SECURITY/20160303/omg_shared_ca_permissions.xsd, a given rule should be constructed in the following order:
- publish tags
- subscribe tags
- relay tags

Currently we interleave publish and subscribe tags as we go

mikaelarguedas:
with the overhaul of permission generation from #72, the generated files now validate the schema

PR #62, opened 2018-07-13
refer to new DDS-Security 1.1 XSDs now that they're available

mikaelarguedas:
Now that the XSDs for DDS_Security 1.1 are out. we can refer to them (our generated files were already using tags from 1.1).

While our "wildcard" generated file comply with the new schema:
```
$ wget https://www.omg.org/spec/DDS-SECURITY/20170901/omg_shared_ca_governance.xsd
$ xmllint -schema omg_shared_ca_governance.xsd ./governance.xml
./governance.xml validates
$ wget https://www.omg.org/spec/DDS-SECURITY/20170901/omg_shared_ca_permissions.xsd
$ xmllint -schema omg_shared_ca_permissions.xsd ./permissions.xml 
./permissions.xml validates
```

The ones containing explicit access control rules still don't as described in #49

### DIFF
--- a/sros2/api/__init__.py
+++ b/sros2/api/__init__.py
@@ -145,17 +145,12 @@ def create_governance_file(path, domain_id):
         f.write("""\
 <?xml version="1.0" encoding="UTF-8"?>
 <dds xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:noNamespaceSchemaLocation="http://www.omg.org/spec/DDS-SECURITY/20160303/omg_shared_ca_governance.xsd">
+    xsi:noNamespaceSchemaLocation="http://www.omg.org/spec/DDS-SECURITY/20170901/omg_shared_ca_governance.xsd">
     <domain_access_rules>
         <domain_rule>
             <domains>
               <id>%s</id>
             </domains>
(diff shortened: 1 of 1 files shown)


sros2/api/__init__.py
```diff
@@ -145,17 +145,12 @@ def create_governance_file(path, domain_id):
         f.write("""\
 <?xml version="1.0" encoding="UTF-8"?>
 <dds xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:noNamespaceSchemaLocation="http://www.omg.org/spec/DDS-SECURITY/20160303/omg_shared_ca_governance.xsd">
+    xsi:noNamespaceSchemaLocation="http://www.omg.org/spec/DDS-SECURITY/20170901/omg_shared_ca_governance.xsd">
```
dirk-thomas: Why not use `https` here since the URL is redirected to that one anyway?

Same below.

mikaelarguedas: Because the examples in the spec and the examples on OMG website use http.
As there is currently no downside in keeping the http link I figured I'd keep it consistent with the examples.
And who knows, maybe OMG plan ons providing an http hosting and not a redirect in the future ? :crossed_fingers:

PR #64, opened 2018-08-17
raise FileNotFoundError if provided permission file doesn't exist

### LINKED ISSUE #41: ros2 security create_permission should abort when a policy file does not exist

mikaelarguedas:
Fixes #41

### DIFF
--- a/sros2/api/__init__.py
+++ b/sros2/api/__init__.py
@@ -425,7 +425,7 @@ def create_permission_file(path, name, domain_id, permissions_dict):
 def get_permissions(name, policy_file_path):
     import yaml
     if not os.path.isfile(policy_file_path):
-        raise RuntimeError("policy file '%s' does not exist" % policy_file_path)
+        raise FileNotFoundError("policy file '%s' does not exist" % policy_file_path)
     with open(policy_file_path, 'r') as graph_permissions_file:
         try:
             graph = yaml.load(graph_permissions_file)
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Is the `FileNotFoundError` exception beind handled by the caller? If not this might result in a stacktrace for the user. On the verb side re-raising this as a `RuntimeError` might be better user experience as it only prints the error message but not a stacktrace.

mikaelarguedas:
It is currently not being handled at the higher level.
I'll just raise a `RuntimeError` with a more detail error message here to keep the change minimal

dirk-thomas:
Since the modules in the `api` directory are supposed to be reused the more specific exception might be better...

mikaelarguedas:
thanks @dirk-thomas for the reviews!

PR #71, opened 2018-11-28
Ros2 service crystal

ross-desmond:
Added generate_permissions file based on node graph api. Enables security of services through sros policy yaml and definition file of how the policy yaml is constructed.

ros2_security_helper enables cmake macro to generate security files through colcon build for development and deployment. Note, this does not get built due to the structure of sros 2. OSRF please let me know where this macro should live to enable security by default when developing on ROS2
e.g:
`find_package(ros2_security_helper REQUIRED)`
`ros2_secure_node(NODES minimal_publisher minimal_subscriber minimal_service)`


mjcarroll:
@ross-desmond I think that we need to get the underlying pull requests in for the node graph API before we can considering getting this merged in to sros2.  Did you make any progress on the other implementations outlined in this comment? https://github.com/ros2/rcl/pull/333#issuecomment-440458945

ross-desmond:
They are on the way and will be available at the latest friday. Ill ensure
all our pr's are at the top of master. Would it be best to ping out at that
time?

mjcarroll:
Yes please!

mjcarroll:
@ruffsl I'm just now getting up to speed on sros2 stuff after transitioning it from Mikael, would you mind putting a second set of eyes on this for me?

ross-desmond:
In the future, is anyone opposed to using jinja for templates?

wjwwood:
> General comment: In the future, is anyone opposed to using jinja for templates?

As opposed to? Or for which task?

clalancette:
> sros2 has all the xml code in the **api**.py, it would be nice to have template files that python inserts the correct data to. This is just a though, it will not change this PR

I won't go too far down this road here, but my concerns would be whether a) jinja is supported on macOS, Windows, and Linux, and b) adding in yet another templating system.  We already have empy (which is non-specific to XML), and I find templating systems difficult to master.  Adding more of them doesn't seem like a recipe for maintainable code down the road.

wjwwood:
> sros2 has all the xml code in the api.py, it would be nice to have template files that python inserts the correct data to. This is just a though, it will not change this PR

That would be fine with me. We do already have a dependency on `EmPy` which is a template engine, so using that would avoid a new dependency, but I don't feel strongly about that. +1 for moving the file text out of the python files (but later though).

ruffsl:
> I won't go too far down this road here, but my concerns would be whether a) jinja is supported on macOS, Windows, and Linux, and b) adding in yet another templating system. We already have empy (which is non-specific to XML), and I find templating systems difficult to master. Adding more of them doesn't seem like a recipe for maintainable code down the road.

+1 for not adding yet another PL specific templating library. I'd suggest we formalize the permission transform to be agnostic of any programming implementation, to be more formally typed, and transparent to promote standardization of how ROS2 permissions are generated for DDS, or any other transport implementation in the future for that matter.

I've created a branch that gives an example using program language agnostic transform for converting ROS2 permission profile markup into transport specific permission XML file. Users could then ecopertate the eXtensible Stylesheet into their own tools to consistently translate the permissions as a known finite transformation. I just spent an hour or so on it, but its relatively simple to support, and I can extend it into this PR if there is interest. See https://github.com/ros2/sros2/pull/72 for current diff.

ross-desmond:
Is this planned for a merge at some point? Is there any help I can provide?

examples/sample_policy.yaml
```diff
@@ -1,27 +1,36 @@
 nodes:
   listener:
     topics:
-      chatter:
-        allow: s # can subscribe to chatter
+      /chatter:
```
jacobperron: Unless it has become a requirement that topic names should start with the global namespace, `'/'`, I don't think this change should be necessary.

examples/policy_definition.txt
```diff
@@ -0,0 +1,30 @@
+## Values used in policy generation
```
jacobperron: This file looks like Markdown, perhaps it should be renamed with the extension `.md` (it would make viewing it in GitHub more palatable anyways).

ross-desmond: Good idea, that can be changed.

sros2/api/__init__.py
```diff
@@ -328,18 +385,20 @@ def create_permission_file(path, name, domain_id, permissions_dict):
...
         # TODO(mikaelarguedas) update dictionary based on existing rule
         # if it already exists (rather than overriding the rule)
-        topic_dict['parameter_events'] = {'allow': 'ps'}
-        topic_dict['clock'] = {'allow': 's'}
+        topic_dict['/parameter_events'] = {'allow': ['publish', 'subscribe']}
+        topic_dict['/clock'] = {'allow': ['subscribe']}
         # we have some policies to add !
         for topic_name, policy in topic_dict.items():
             tags = []
-            if policy['allow'] == 'ps':
```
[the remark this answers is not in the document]
ross-desmond: It would help keep the PR succinct, but then there would be different ways of describing the security for services(-request, -response) vs topics (ps). It makes the parsing and description of the elements more verbose and robust.

sros2/api/__init__.py
```diff
@@ -352,12 +411,10 @@ def create_permission_file(path, name, domain_id, permissions_dict):
             <topic>%s</topic>
           </topics>
         </%s>
-""" % (tag, 'rt/' + topic_name, tag)
+""" % (tag, 'rt' + topic_name, tag)
```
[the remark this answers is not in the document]
ross-desmond: We can add a check

examples/sample_policy.yaml
ross-desmond: Private topic names will be specified from global with the entire path to avoid confusion in the current revision. Is this something that is not desired?

ross-desmond:
Thank you for the review @jacobperron.

Just a reminder the security_helpers folder in sros2 may want to be split out to another package as it does not build under the current sros2 package structure. If we believe this helper macro should stay in sros2 then I'll need to restructure to build sros2 python modules and the helper macro, thoughts?

examples/sample_policy.yaml
ross-desmond: It can either have the / or not, but they will be globally referenced

jacobperron:
>Just a reminder the security_helpers folder in sros2 may want to be split out to another package as it does not build under the current sros2 package structure. If we believe this helper macro should stay in sros2 then I'll need to restructure to build sros2 python modules and the helper macro, thoughts?

IMO, the cmake package can remain in this repo (but remain a separate package) as it matches the repo description. However, I would consider remaining the package to something like `sros2_cmake`, since its main purpose is providing cmake macros for interacting with ROS 2 security.
I would restructure this repo so that the two packages are at top along with the general docs about sros2:

```
sros2
  |_ sros2 (python lib)
  |_ test
  |_ package.xml
  |_ setup.py
  |_ ... etc
sros2_cmake
  |_ cmake
    |_ ros2_create_keystore.cmake
    |_ ros2_secure_node.cmake
  |_ CMakeLists.txt
  |_ package.xml
  |_ ... etc
CONTRIBUTING.md
LICENSE
README.md
SROS2_Linux.md
SROS2_MacOS.md
SROS2_Windows.md
```

I'd like to get some other thoughts on this proposal.

If we go this approach, I think it would be better if the cmake package is added in a separate PR as it is distinct from the CLI changes and is a larger change.

CHANGELOG.rst
```diff
@@ -2,6 +2,19 @@
 Changelog for package sros2
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+0.6.0 (2018-12-07)
```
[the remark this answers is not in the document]
ross-desmond: This PR was merged with master is all, removed

ross-desmond:
Fixing flake8 errors, incoming

mjcarroll:
CI up to SROS2: 

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6003/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2512/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4956/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5880/)

ross-desmond:
> > Just a reminder the security_helpers folder in sros2 may want to be split out to another package as it does not build under the current sros2 package structure. If we believe this helper macro should stay in sros2 then I'll need to restructure to build sros2 python modules and the helper macro, thoughts?
> 
> IMO, the cmake package can remain in this repo (but remain a separate package) as it matches the repo description. However, I would consider remaining the package to something like `sros2_cmake`, since its main purpose is providing cmake macros for interacting with ROS 2 security.
> I would restructure this repo so that the two packages are at top along with the general docs about sros2:
> 
> ```
> sros2
>   |_ sros2 (python lib)
>   |_ test
>   |_ package.xml
>   |_ setup.py
>   |_ ... etc
> sros2_cmake
>   |_ cmake
>     |_ ros2_create_keystore.cmake
>     |_ ros2_secure_node.cmake
>   |_ CMakeLists.txt
>   |_ package.xml
>   |_ ... etc
> CONTRIBUTING.md
> LICENSE
> README.md
> SROS2_Linux.md
> SROS2_MacOS.md
> SROS2_Windows.md
> ```
> 
> I'd like to get some other thoughts on this proposal.
> 
> If we go this approach, I think it would be better if the cmake package is added in a separate PR as it is distinct from the CLI changes and is a larger change.

I'm in for this, I'll remove it from this package, @ruffsl  how does that sound?

ruffsl:
> I'm in for this, I'll remove it from this package, @ruffsl how does that sound?

I'd be fine with that restructuring. I'd still like to recommend the use of stylesheets to compile the xml permissions, rather than a yaml implementation specific approach. I'd welcome feedback on https://github.com/ros2/sros2/pull/72 , and could rebase if the repo is restructured as above.

ross-desmond:
> > I'm in for this, I'll remove it from this package, @ruffsl how does that sound?
> 
> I'd be fine with that restructuring. I'd still like to recommend the use of stylesheets to compile the xml permissions, rather than a yaml implementation specific approach. I'd welcome feedback on #72 , and could rebase if the repo is restructured as above.

Awesome, I'm happy to move to XML as well. In the meantime, using generate_permissions with create_permission will have the same result as with json and xml. Let's get the xml in for the next patch release?

PR #72, opened 2018-12-05
Use XML and XSLT to perform permission transform

ruffsl:
This profile is language agnostic, and transforms can be added for more transports.

TODO:
 - [x] correct transform for actions to properly reflect current mapping in crystal release
 - [x] modify ros2 cli to use template to convert permissions
 - [x] add xsd schema for profile xml file format
 - [x] use xsd schemas for checking for both input profile and output permissions
 - [x] use ROS_DOMAIN env to set domain id during transform
 - [x] support relative, global, and private namespace resolving
 - [x] support nested including for modulare profiles and policy reuse
 - [x] support profile generation from monitoring currently running nodes


ruffsl:
So at this point, this xml profile format is working in generating complaint permission files. Before I go an add some more tests, perhaps folks would like to give some feedback on the profile structure, or play around with making profiles. The commands work the same way, only it takes in an xml file now for generating permissions, and the node name given should be FQN: e.g `/talker` instead of `talker`.

This should allow the permission generation for a given transport to be less coupled with the heigherleve API and CLI. We can add more transforms for future transports and version them as well to support multiple versions/releases of the same transport.

ping @jacobperron @mjcarroll and perhaps @ross-desmond.

package.xml
```diff
@@ -13,6 +13,7 @@
   <depend>ros2cli</depend>
 
   <exec_depend>openssl</exec_depend>
+  <exec_depend>python3-lxml</exec_depend>
```
jacobperron: AFAICT, there is not rosdep rule for `python3-lxml`:

```
sros2: Cannot locate rosdep definition for [python3-lxml]
```

ruffsl: I'm not sure how to add a package key to the ros2/rosdistro listing as one would normally do for ros1, as there doesn't seem to be a `rosdep/python.yaml` in the ros2 fork given the structure has changed. How would I go about doing the ros2 equivalent of this: https://github.com/ros/rosdistro/pull/12314

python3-lxml is already released in the official ubuntu, debian, and fedora distros:  
* https://packages.ubuntu.com/bionic/python3-lxml
* https://packages.debian.org/stretch/python3-lxml
* https://apps.fedoraproject.org/packages/python3-lxml

jacobperron: We're still using the ROS 1 repo for rosdep: https://github.com/ros/rosdistro/tree/master/rosdep
You can add the key there.

ruffsl: @jacobperron , please see https://github.com/ros/rosdistro/pull/19824

sros2/policy/templates/dds/policy.xsl
```diff
@@ -0,0 +1,207 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<xsl:stylesheet version="1.0"
+xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
+
+<xsl:variable name="templace_validity">
```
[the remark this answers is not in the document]
ruffsl: Fixed in 9a12dc5

examples/profile_to_permissions.py
```diff
@@ -0,0 +1,20 @@
+import lxml.etree as ET
+
+# Set paths
+profile_xml_path = 'sample_policy.xml'
+profile_xslt_path = 'profile_template.xslt'
```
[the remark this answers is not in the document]
ruffsl: Updated minimal working example to use sros2 package data in 2f2c1b8

ruffsl:
@jacobperron , I've updated the key generation, as well as the governance templating, so I no longer encounter there error above you've show. Please give it another try, just be sure to provide the names of the nodes as a FQN with a root slash:

``` shell
ros2 security create_keystore demo_keys
ros2 security create_key demo_keys /talker
ros2 security create_key demo_keys /talker
ros2 security create_permission demo_keys /talker  <path_to>/sample_policy.xm
```

You should also be able to create keys for nodes nested under deeping namespaces:

``` shell
ros2 security create_key demo_keys /foo/talker
ros2 security create_key demo_keys /foo/bar/talker
ros2 security create_permission demo_keys /foo/bar/talker  <path_to>/sample_policy.xm
```

Note: you'd have to modify the sample_policy.xml example to include the `ns=/foo/bar` attribute so that sros command may find the matching policy.

ruffsl:
ping @jacobperron @mjcarroll and perhaps @ross-desmond
Please see the updated example, it demonstrates all the different forms of private, relative, and global namespaces of permissions, as well a composable primitives via recursive includes. 

Services and actions for Crystal are also ported.

ruffsl:
ping @jacobperron @mjcarroll , could I get some feedback or CI triggered? This is ready for review..

jacobperron:
> ping @jacobperron @mjcarroll , could I get some feedback or CI triggered? This is ready for review..

Sorry for the delay.
I'll get to a review soon. Here's CI in the mean time:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6008/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2515/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4959/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5883/)

package.xml
ruffsl: Does `python3-lxml` also need to be a `<test_depend>` or just a `<depend>`?

jacobperron: No, I don't think so. According to http://wiki.ros.org/catkin/package.xml#Dependencies:

> They should never duplicate any dependencies already mentioned as build or run dependencies.

jacobperron:
@ruffsl I guess our CI doesn't install deps based on the package.xml. I'll look into it.

mjcarroll:
Currently, it does not, it's based on a fixed set of packages.  Many deps are installed in this neighborhood: https://github.com/ros2/ci/blob/master/linux_docker_resources/Dockerfile#L106

If we can get a branch on `ros2/ci` that includes that package, then we can run a CI job against that.  (Doesn't fix linux or macOS).

mjcarroll:
Check that, we could probably inject it here for all platforms: https://github.com/ros2/ci/blob/master/ros2_batch_job/__main__.py#L49

ruffsl:
I think you'll have to fix the CI branch again. https://github.com/ros2/ci/commit/579067951b30910d09b8b007cc63c22600671f6b#r31903461

ruffsl:
ping @jacobperron @mjcarroll and perhaps @ross-desmond

I've just refactored the PR to the new repo structure and have updated the generate_profile verb to create profiles using the new xml format. Please take a look.

SROS2_Windows.md
```diff
@@ -124,19 +124,19 @@ ros2 run demo_nodes_cpp talker __node:=not_talker
...
-curl -k https://raw.githubusercontent.com/ros2/sros2/master/examples/sample_policy.yaml -o .\demo_keys\policies.yaml
+curl -k https://raw.githubusercontent.com/ros2/sros2/master/examples/sample_policy.xml -o .\demo_keys\policy.xml
 ```
 
 And now we will use it to generate the XML permission files expected by the middleware:
 
 ```bat
-ros2 security create_permission demo_keys talker demo_keys/policies.yaml
-ros2 security create_permission demo_keys listener demo_keys/policies.yaml
+ros2 security create_permission demo_keys /talker demo_keys/policy.xml
```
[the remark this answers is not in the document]
ruffsl: What is a good way to store and locate a ros2 package asset via command line?
https://answers.ros.org/question/288501/ros2-equivalent-of-rospackagegetpath/

This close, but not quite it:
```
ros2 security create_permission demo_keys /talker $(ros2 pkg prefix sros2)/test/policies/sample_policy.xml
```

sros2/setup.py
```diff
@@ -1,6 +1,22 @@
...
+def package_files(directory):
+    paths = []
+    for (path, directories, filenames) in os.walk(directory):
+        for filename in filenames:
+            paths.append(os.path.join('..', path, filename))
+    return paths
+
+
+extra_files = []
+extra_files.extend(package_files('sros2/policy/schemas'))
```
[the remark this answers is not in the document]
ruffsl: Done

SROS2_Linux.md
```diff
@@ -172,19 +172,19 @@ ros2 run demo_nodes_py listener
...
 To increase the level of security in the system, you can define strict limits, known as access control, which restrict what each node is able to do.
 For example, one node would be able to publish to a particular topic, and another node might be able to subscribe to that topic.
-To do this, we will use the sample policy file provided in `examples/sample_policy.yaml`.
+To do this, we will use the sample policy file provided in `examples/sample_policy.xml`.
 
 First, we will copy this sample policy file into our keystore:
 
 ```bash
-curl -sk https://raw.githubusercontent.com/ros2/sros2/master/examples/sample_policy.yaml -o ./demo_keys/policies.yaml
+curl -sk https://raw.githubusercontent.com/ros2/sros2/master/examples/sample_policy.xml -o ./demo_keys/policy.xml
```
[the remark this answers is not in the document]
ruffsl: When with using svn to download the necessary subdirectory. E.g:
https://stackoverflow.com/questions/7106012/download-a-single-folder-or-directory-from-a-github-repo

SROS2_Windows.md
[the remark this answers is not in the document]
ruffsl: I updated the readme to use the simple_example.xml instead. The `default.xml` is only used to bootstrap the permission before any policy profile is specified. It simple allows for all ros interfaces, yet is more conservative than simple `<topic>*</topic>`.

mjcarroll:
@ruffsl I asked a few days ago, but it appears lost to the internet.  Are you trying to land this in the patch release?

ruffsl:
> @ruffsl I asked a few days ago, but it appears lost to the internet. Are you trying to land this in the patch release?

That would be nice. I'd like to get the profile format out there and get some feedback from real users.

I'd like to eventually overhaul the security CLI, as I don't think the positional arguments are intuitive, and we should refactor the crypto to use python APIs rather than openssl via subrosses, buth those are out of scope for the current PR.

@mjcarroll , did you figure out how to get the CI to use the package.xml for dependency resolution?

mjcarroll:
I think the CI changes are going to make it not fit in the patch release, but since this doesn't have a huge downstream impact, we can probably slip it into the next sync.

ruffsl:
> but since this doesn't have a huge downstream impact, we can probably slip it into the next sync.

True, I haven't yet found anything that used the past profile templates, as they where previously incomplete even for the simple demo talker case. 

If this PR currently looks good for someone other than myself, then let's go for that sync.

ruffsl:
@mjcarroll , any progress with fixing the CI?

mjcarroll:
Sorry @ruffsl still working this, I think I have a solution for Windows and Linux, coordinating with @nuclearsandwich on what a macOS fix is.

ruffsl:
Thanks @mjcarrol , more things seem to be bunching up here waiting on this PR.

mjcarroll:
Got Linux/macOS/Windows working, I believe:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6226/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2665/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5127/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6071/)

mjcarroll:
Aarch64 with libxml2-dev and libxslt-dev installed (no wheel available, so must be built from source)

* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2667/)

mjcarroll:
Nevermind, lxml takes a while to build on aarch64 (for no real benefit), just using the apt packages and skipping the pip install on Ubuntu: 

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6229/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2670/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5130/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6074/)

mjcarroll:
Just need to get https://github.com/ros2/ci/pull/246 merged, then this should be good to go!

mjcarroll:
@ruffsl, @sloretz brought to my attention that `lxml` is GPL licensed.  Is that an issue in this case?  It appears to only being used to transform/manipulate policy, but it may have an effect on the package license.

Thoughts?

mjcarroll:
False alarm! BSD: https://github.com/lxml/lxml/blob/master/LICENSES.txt (but test code is GPL)

package.xml
sloretz: Can the built in `xml` module be used instead of `lxml`? https://docs.python.org/3.6/library/xml.etree.elementtree.html

ruffsl: > Can the built in `xml` module be used instead of `lxml`?

If all we needed was to parse XML markup, then yes the standard python module `xml` would suffice. However in the interest of keeping the templating pipeline implementation agnostic (so that SROS2 security profiles may be consumed by other tools and languages), I've elected to use XML templating as well. This necessitates the use of both schema validation and template processing for the complete profile-to-permission pipeline. There are few python modules that provide XSD validation, and fewer still that also support XST templating. At the moment, `lxml` seems to be the only game in town.

https://stackoverflow.com/questions/299588/validating-with-an-xml-schema-in-python

For a while with a previous work, I was using `xmlschema`; a python native XSD engine (lacks a XST processor). Yet I was somewhat misusing it as a XST engine as well for pipelining permission translation; something a that proper XST files are much more adapt as solving.

https://pypi.org/project/xmlschema/

Also, I'm not sure if the latest fix to the standard modul's `Xincludes` processor as been backported to the version of python we target:

https://bugs.python.org/issue20928 

Thus the case for `lxml`.

ruffsl:
> just using the apt packages and skipping the pip install on Ubuntu:

Yea, I was about to suggest that, given its already packaged by most distros:  
https://github.com/ros/rosdistro/pull/19824

> It appears to only being used to transform/manipulate policy, but it may have an effect on the package license.
> False alarm! BSD

Yea, might have been an issue, but fortunately `lxml` itself is shipped under a BSD license:  
https://lxml.de/index.html#license

ruffsl:
All tests are [Green](https://github.com/ros2/sros2/pull/72#issuecomment-465202300)!  
Please make this PR Purple!

mjcarroll:
@ruffsl is this still ready to merge after that force push?  Just got the LXML dep in.

ruffsl:
I rebased wrt master, as I misread a remark elsewhere. Suppose I didn't need to, my bad. I can push the previous branch back if you'd like. Nothing's changed though.

mjcarroll:
Nope just verifying!  Thanks, and sorry that it has taken so long.

PR #74, opened 2019-01-11
SROS2 cmake restructure

ross-desmond:
Restructures sros2 to the following: @jacobperron 

> > Just a reminder the security_helpers folder in sros2 may want to be split out to another package as it does not build under the current sros2 package structure. If we believe this helper macro should stay in sros2 then I'll need to restructure to build sros2 python modules and the helper macro, thoughts?
> 
> IMO, the cmake package can remain in this repo (but remain a separate package) as it matches the repo description. However, I would consider remaining the package to something like `sros2_cmake`, since its main purpose is providing cmake macros for interacting with ROS 2 security.
> I would restructure this repo so that the two packages are at top along with the general docs about sros2:
> 
> ```
> sros2
>   |_ sros2 (python lib)
>   |_ test
>   |_ package.xml
>   |_ setup.py
>   |_ ... etc
> sros2_cmake
>   |_ cmake
>     |_ ros2_create_keystore.cmake
>     |_ ros2_secure_node.cmake
>   |_ CMakeLists.txt
>   |_ package.xml
>   |_ ... etc
> CONTRIBUTING.md
> LICENSE
> README.md
> SROS2_Linux.md
> SROS2_MacOS.md
> SROS2_Windows.md
> ```
> 
> I'd like to get some other thoughts on this proposal.
> 
> If we go this approach, I think it would be better if the cmake package is added in a separate PR as it is distinct from the CLI changes and is a larger change.

Depends on https://github.com/ros2/sros2/pull/71


ruffsl:
Could you separate the migration of the sros2 python lib folder rename/restructure vs the addition of the sros2_cmake stuff into seperate PRs? Then I could work on rebasing my xml PR sooner while the cmake stuff is still in review.

PR #80, opened 2019-02-18
add logging permission profile

mikaelarguedas:
This PR targets #72, not master

Since https://github.com/ros2/rcl/pull/350 all node need publish access to `/rosout` by default.

I called it logging in case we have other related default logging topics in the future.

@ruffsl FYI

### DIFF
--- a/sros2/test/policies/common/node/logging.xml
+++ b/sros2/test/policies/common/node/logging.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <profile>
   <topics publish="ALLOW" >
-    <topic>/rosout</topic>
+    <topic>rosout</topic>
   </topics>
 </profile>
--- a/sros2/test/policies/common/node/logging.xml
+++ b/sros2/test/policies/common/node/logging.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<profile>
+  <topics publish="ALLOW" >
+    <topic>/rosout</topic>
+  </topics>
+</profile>
--- a/sros2/test/policies/common/node.xml
+++ b/sros2/test/policies/common/node.xml
@@ -1,5 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <profile xmlns:xi="http://www.w3.org/2003/XInclude">
+  <xi:include href="node/logging.xml"
+    xpointer="xpointer(/profile/*)"/>
   <xi:include href="node/time.xml"
     xpointer="xpointer(/profile/*)"/>
   <xi:include href="node/parameters.xml"
--- a/sros2/test/policies/permissions.xml
+++ b/sros2/test/policies/permissions.xml
@@ -26,6 +26,7 @@
             <topic>rr/talker/set_parameters_atomicallyReply</topic>
             <topic>rt/chatter</topic>
             <topic>rt/parameter_events</topic>
+            <topic>rt/rosout</topic>
           </topics>
         </publish>
         <subscribe>
@@ -74,6 +75,7 @@
             <topic>rr/listener/set_parametersReply</topic>
             <topic>rr/listener/set_parameters_atomicallyReply</topic>
             <topic>rt/parameter_events</topic>
+            <topic>rt/rosout</topic>
           </topics>
         </publish>
         <subscribe>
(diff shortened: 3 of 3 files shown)


ruffsl:
Ah man... its like ROS1 all over again. Why is everything is needing to publish to root level topics and parameters by default.

The test case may need to be updated though:

https://github.com/ros2/sros2/pull/72/files#diff-9534083787043b52ef6d287331dfc69bR53

sros2/test/policies/common/node/logging.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<profile>
+  <topics publish="ALLOW" >
+    <topic>/rosout</topic>
```
ruffsl: Should this be `<topic>/rosout</topic>`, `<topic>~rosout</topic>` or just `<topic>rosout</topic>`?
I.e. should the topic be left as global, private or relative?

I haven't tested yet, but it looked like from the PR you link that `rosout` might only be needed:

https://github.com/ros2/rcl/pull/350/files#diff-bf8aead5abb376e0ca7e38051163fb9dR34

mikaelarguedas: > I haven't tested yet, but it looked like from the PR you link that `rosout` might only be needed:

I actually didnt look in detail at the PR, this change made the access control tutorial work for me while I couldn't get the one on master to work.

> Should this be `<topic>/rosout</topic>`, `<topic>~rosout</topic>` or just `<topic>rosout</topic>`?

Good question, maybe reviewers of  [ros2/rcl#350](https://github.com/ros2/rcl/pull/350) can provide guidance here (@jacobperron @nuclearsandwich)

mikaelarguedas:
> The test case may need to be updated though:

Good point :+1:, test updated in https://github.com/ros2/sros2/pull/80/commits/e385cbd9ae806ddab06bcab92adbf96f6c25ba50

sros2/test/policies/common/node/logging.xml
jacobperron: If we namespace a node, e.g.

        ros2 run examples_rclcpp_minimal_publisher publisher_member_function __ns:=/foo

The `rosout` topic does not get pushed into the namespace. So I think the global namespace for `/rosout` makes sense in this context.
Although, I'm not sure if this is intended behavior. I couldn't find any explicit discussion, for or against, keeping a global `/rosout` topic for all nodes (https://discourse.ros.org/t/ros2-logging/6469).

ruffsl: Ok, if it doesn't remap, then leaving it global is applicable.

BTW, @jacobperron , if `ROSOUT_TOPIC_NAME` is left unqualified, what keeps that topic from being remapped? Shouldn't `ROSOUT_TOPIC_NAME` be `/rosout`?

https://github.com/ros2/rcl/pull/350/files#diff-bf8aead5abb376e0ca7e38051163fb9dR166

jacobperron: That's a good question. Looking at the rosout implementation, it isn't immediately obvious to me why `rosout` is not being remapped...

jacobperron: Nevermind. I mistakenly read the output of `ros2 topic list`. `rosout` **does** get remapped in my example to `/foo/rosout`.

mikaelarguedas: > Nevermind. I mistakenly read the output of `ros2 topic list`. `rosout` **does** get remapped in my example to `/foo/rosout`.

Good to know, updated accordingly in https://github.com/ros2/sros2/pull/80/commits/f2061f118e018d25fc5dc461445b60b05be826bc

jacobperron:
Ignore unrelated test failures in `rclcpp_action` and `test_cli`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6248/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2683/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5143/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6092/)

jacobperron:
Looks like the changes in [ros2/ci#246](https://github.com/ros2/ci/pull/246) need to be rebased on master, I'll just wait until it is merged before triggering CI.
@ruffsl I think you can go ahead and merge this into your branch if you're happy with it.

PR #83, opened 2019-02-21
Correct sros2 cli test folder location

ruffsl:
I'm not sure how only these rename and rm commits got dropped from my rebasing of #72 .
These are the only changes that had rebase differ from the PR commits prior to rebasing.


mikaelarguedas:
lgtm, tests work for me with that branch :+1:

Will there be a document replacing [policy_definition.md](https://github.com/ros2/sros2/pull/83/files#diff-4dd86053375383c81c1533ed3d87d7cb)?

ruffsl:
> Will there be a document replacing policy_definition.md

I won't have time until the end of this month, but feel free to take a stab at it. Having a second person combe through the schema and templates would help in case I missed a corner case.

mikaelarguedas:
> I won't have time until the end of this month, but feel free to take a stab at it.

Sounds good, I'll see if I find time to look into it.
Do you think this document is a requirement for the xml based policy to be part of the next Crystal patch release?

ruffsl:
> Do you think this document is a requirement for the xml based policy to be part of the next Crystal patch release?

I'm not sure it would be a requirement, given the examples in the test folder are a bit self explanatory, but something short explaining the features of the namespacing parameters and permission flags might be helpful.

https://github.com/ros2/sros2/blob/master/SROS2_Linux.md#access-control

PR #84, opened 2019-02-21
add reference to schema in generated permission files

mikaelarguedas:
This PR adds a reference to the schema used in the generated xml files (similar to what is in the [default governance file](https://github.com/ros2/sros2/blob/79fffad94045e60edb9f3c2e7c6fe614bf1d1f3e/sros2/sros2/policy/defaults/dds/governance.xml#L2-L3)) so that users can perform validation on the files.

@ruffsl It looks like this repo now has a copy of the permission and governance XSDs from the OMG website. Is the expected way of managing schema to commit new copies of the schemas to this repo  when a new DDS Security version comes out?

### DIFF
--- a/sros2/sros2/policy/templates/dds/permissions.xsl
+++ b/sros2/sros2/policy/templates/dds/permissions.xsl
@@ -22,7 +22,8 @@
 
 <xsl:template match="/policy/profiles">
   <xsl:variable name="dds">
-    <dds>
+    <dds xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:noNamespaceSchemaLocation="http://www.omg.org/spec/DDS-SECURITY/20170901/omg_shared_ca_permissions.xsd">
       <permissions>
         <xsl:for-each select="profile">
           <xsl:variable name="_ns">


ruffsl:
> Is the expected way of managing schema to commit new copies of the schemas to this repo when a new DDS Security version comes out?

I'd expect so. I don't know of another way the CLI could acquire the DDS schema without it being pre-installed, other than requiring users have constant internet access to the public internet :earth_asia: . Can we get it from the vendor installed?

mikaelarguedas:
> Can we get it from the vendor installed?

I remember that at least RTI and Fast-RTPS provide a version of these, not sure about Opensplice or other vendors.
Potential alternatives:
- Use the one provided by the vendors (though it assumes they provide the same version, which was not true in the past)
- Fetch the schema  from the OMG website at sros2 build time and install it, this way we dont need to embed a copy here or a valid internet connexion when using the CLI (though we'll likely need to update the download URL when a new version comes out).
- Host them on `http://download.ros.org` so that they have an http URL (some tools like xmllint can't fetch https hosted schemas), this could potentially allow us to keep the same URL.

I have nothing against the current solution about embedding it in the repo, just brainstorming if we could come up with anything avoiding duplication

ruffsl:
> Fetch the schema  from the OMG website at sros2 build time and install it, this way we dont need to embed a copy here or a valid internet connexion when using the CLI (though we'll likely need to update the download URL when a new version comes out).

That sounds nice! Could we add a bit of logic to the setup.py to first check the source folder for it before downloading a version from OMG as a fall back. This would keep it easy for developers to use custom schemas, or test unreleased schemas in beta.

> some tools like xmllint can't fetch https hosted schemas

The fact the no XML template processor can use HTTPS was infuriating when I was prototyping! If we do scrape the XSD from OMG at build time, lets makes sure if via https though.

> I have nothing against the current solution about embedding it in the repo, just brainstorming if we could come up with anything avoiding duplication

The XSD hasn't changed often TBH, I think the latest release merely added and optional `<script>` element that I couldn't find in the documentation.

mikaelarguedas:
Pulling it at build time sounds good to me. This will likely be a separate PR though as this one is just for consistency accross governance/permission xml files and allow end consumers to have a self-contained permission files.

I would advocate to get this in as is as I don't know if/when I'll have time to work on the next step.

> The XSD hasn't changed often TBH, I think the latest release merely added and optional `<script>` element that I couldn't find in the documentation.

Yeah last change I remember was the renaming of `allow_unauthenticated_join` vs `allow_unauthenticated_participants`.

ruffsl:
> I would advocate to get this in as is as I don't know if/when I'll have time to work on the next step.

Feel free to merge when ready. Could you open a ticket for this so we don't forget.

mikaelarguedas:
> Feel free to merge when ready. Could you open a ticket for this so we don't forget.

:+1: https://github.com/ros2/sros2/issues/86

PR #85, opened 2019-02-22
[ros2_cmake] fix status print to match commands invoked

mikaelarguedas:
The status message printed for `create_key` does not match the command actually ran (additional `${policy}` argument making the command fail with `ros2: error: unrecognized arguments: foo.xml`.)

This PR uses the same string for the command and the message to avoid discrepancies in the future

---

The style of this file doesn't seem to match the [ROS 2 style guide for CMake](https://index.ros.org/doc/ros2/Contributing/Developer-Guide/#cmake). @mjcarroll @ross-desmond could we add linters to this package?


jacobperron:
Added lint test in #90.

PR #81, opened 2019-02-18
[Issue 79] Fix permissions tutorial

ross-desmond:
**Summary**
* SROS2 assumed services were included in the policy yaml file
* Format errors in the sample_policy prevented yaml parse

-----------

SROS2 is moving to xml, if a fix is required in the yaml version, here it is.

Resolves #79

Issue #79, opened 2019-02-18
Tutorial permissions not working


mikaelarguedas:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu Bionic
- Installation type:
  - source
- Version or commit hash:
  - : sros2 https://github.com/ros2/sros2/commit/f145bf9f632d1ec2752575fce79843d02b1da76d
- DDS implementation:
  - N/A
- Client library (if applicable):
  - N/A

#### Steps to reproduce issue

Follow the SROS2_Linux Tutorial until [Access control section](https://github.com/ros2/sros2/blob/master/SROS2_Linux.md#access-control)

or
```
$ docker run -it --rm ros:crystal
apt update && apt upgrade -y && apt install -y curl ros-crystal-demo-nodes-cpp ros-crystal-demo-nodes-py ros-crystal-sros2 
mkdir ~/sros2_demo && cd ~/sros2_demo
ros2 security create_keystore demo_keys
ros2 security create_key demo_keys talker
ros2 security create_key demo_keys listener
curl -sk https://raw.githubusercontent.com/ros2/sros2/master/examples/sample_policy.yaml -o ./demo_keys/policies.yaml
ros2 security create_permission demo_keys talker demo_keys/policies.yaml
```

#### Expected behavior
Generate permission.xml files and be able to use access control
#### Actual behavior
Fails to create permission files.

#### Additional information
<details><summary>Error1</summary>

```
Traceback (most recent call last):
  File "/opt/ros/crystal/bin/ros2", line 11, in <module>
    load_entry_point('ros2cli==0.6.3', 'console_scripts', 'ros2')()
  File "/opt/ros/crystal/lib/python3.6/site-packages/ros2cli/cli.py", line 69, in main
    rc = extension.main(parser=parser, args=args)
  File "/opt/ros/crystal/lib/python3.6/site-packages/sros2/command/security.py", line 37, in main
    return extension.main(args=args)
  File "/opt/ros/crystal/lib/python3.6/site-packages/sros2/verb/create_permission.py", line 44, in main
    success = create_permission(args)
  File "/opt/ros/crystal/lib/python3.6/site-packages/sros2/api/__init__.py", line 522, in create_permission
    permissions_dict = get_permissions(name, policy_file_path)
  File "/opt/ros/crystal/lib/python3.6/site-packages/sros2/api/__init__.py", line 500, in get_permissions
    return graph['nodes'][name]
KeyError: 'nodes'
```

[...]

PR #93, opened 2019-03-08
[crystal-backport] Backport XML and XSLT functionality

mjcarroll:
Based on discussion in #81 and #79, we can backport this new functionality into Crystal.


mjcarroll:
This reveals that I clearly didn't squash-merge something in the past that I should have :grimacing:

PR #90, opened 2019-03-04
Add CMake lint test to sros2_cmake

jacobperron:
Fixed lint errors accordingly.

PR #95, opened 2019-03-10
linter invocation fixup

mikaelarguedas:
Follow up of #90 

- First commit removes the (apparently) unnecessary `ament_cmake_test` find_package call.
- The second commit removes the dependency on `ament_lint_auto` altogether and only depends on `ament_cmake_lint_cmake`.

Second commit is more subjective, as this is not an ament package, each linter has to be called explicitly. At that point it seems that relying on `ament_lint_auto` loses its benefits while adding dependencies and CMake code. As this package is meant to only provide pure cmake macros it looks like `ament_lint_cmake` is the only linter it will need going forward .

Happy to revert the second commit if the ament_lint_auto approach is preferred.

### DIFF
--- a/sros2_cmake/CMakeLists.txt
+++ b/sros2_cmake/CMakeLists.txt
@@ -8,8 +8,7 @@ set(INCLUDE_INSTALL_DIR include/)
 set(SYSCONFIG_INSTALL_DIR share/${PROJECT_NAME})
 
 if(BUILD_TESTING)
-  find_package(ament_lint_auto REQUIRED)
-  ament_lint_auto_find_test_dependencies()
+  find_package(ament_cmake_lint_cmake REQUIRED)
   ament_lint_cmake()
 endif()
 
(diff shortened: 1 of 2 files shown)


sros2_cmake/CMakeLists.txt
```diff
@@ -7,10 +7,8 @@ set(LIB_INSTALL_DIR lib/)
 set(INCLUDE_INSTALL_DIR include/)
 set(SYSCONFIG_INSTALL_DIR share/${PROJECT_NAME})
 
-find_package(ament_cmake_test REQUIRED)
```
jacobperron: This command sets the build `BUILD_TESTING` flag.
If we remove it, we should do something else to enable testing (I think `enable_testing()`).

dirk-thomas: There is much more logic to it than just calling `enable_testing()`: see https://github.com/ament/ament_cmake/blob/207a07b72af1e5a780887d7c2268df57ef456af3/ament_cmake_test/ament_cmake_test-extras.cmake#L17-L37 so I don't think duplicating that logic makes sense.

mikaelarguedas: https://github.com/ros2/sros2/pull/95/commits/e4ee571a5f41fcff8c84b064c8e3d33624b44aa2 applies suggestion from https://github.com/ros2/sros2/pull/90#discussion_r264031986

PR #104, opened 2019-04-01
complete xml and not yaml files for create_permission

mikaelarguedas:


mikaelarguedas:
> Is there any coverage on this that would justify running CI?

Not that I know of, I believe there is no coverage for autocompletion. Regarding the doc change, there may be flake8 checking the docstring format

mikaelarguedas:
> Not that I know of, I believe there is no coverage for autocompletion. Regarding the doc change, there may be flake8 checking the docstring format

FWIW flake8 passes locally and fails if I remove the trailing dot in the docstring.
So I would say this can be merged as is

PR #117, opened 2019-06-05
Guard against invalid key names

### LINKED ISSUE #113: Guard against creating keys for empty and whitespace only names
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - Debian pkgs
- Version or commit hash:
  - 0.7.0
- DDS implementation:
  - Connext
  - N/A

#### Steps to reproduce issue

Create a keystore:
```
ros2 security create_keystore /tmp/ks
```

Try creating a key with an empty name:
```
ros2 security create_key /tmp/ks /
```

Try creating a key with a name containing only whitespace
```
ros2 security create_key /tmp/ks "/ "
```

#### Expected behavior

Graceful error messages in both cases.

#### Actual behavior

The empty name results in the following output:
```
creating key for identity: '/'
Traceback (most recent call last):
  File "/opt/ros/dashing/bin/ros2", line 11, in <module>
    load_entry_point('ros2cli==0.7.0', 'console_scripts', 'ros2')()
  File "/opt/ros/dashing/lib/python3.6/site-packages/ros2cli/cli.py", line 69, in main
    rc = extension.main(parser=parser, args=args)
  File "/opt/ros/dashing/lib/python3.6/site-packages/sros2/command/security.py", line 37, in main
    return extension.main(args=args)
  File "/opt/ros/dashing/lib/python3.6/site-packages/sros2/verb/create_key.py", line 34, in main
    success = create_key(args.ROOT, args.NAME)
  File "/opt/ros/dashing/lib/python3.6/site-packages/sros2/api/__init__.py", line 439, in create_key
    shutil.copyfile(keystore_governance_path, dest_governance_path)
  File "/usr/lib/python3.6/shutil.py", line 104, in copyfile
    raise SameFileError("{!r} and {!r} are the same file".format(src, dst))
shutil.SameFileError: '/tmp/ks/governance.p7s' and '/tmp/ks/./governance.p7s' are the same file
```

The whitespace name succeeds. I don't think this should be a valid name.

jacobperron:
In particular, guard against keys that only consist of whitespace and '/' characters.
The error message is also improved slightly.

Fixes #113

### DIFF
--- a/sros2/package.xml
+++ b/sros2/package.xml
@@ -10,6 +10,7 @@
   <author email="morgan@osrfoundation.org">Morgan Quigley</author>
   <author>Mikael Arguedas</author>
 
+  <depend>rclpy</depend>
   <depend>ros2cli</depend>
 
   <exec_depend>openssl</exec_depend>
--- a/sros2/sros2/api/__init__.py
+++ b/sros2/sros2/api/__init__.py
@@ -22,6 +22,11 @@ import sys
 
 from lxml import etree
 
+from rclpy.exceptions import InvalidNamespaceException
+from rclpy.exceptions import InvalidNodeNameException
+from rclpy.validate_namespace import validate_namespace
+from rclpy.validate_node_name import validate_node_name
+
 from sros2.policy import (
     get_policy_default,
     get_transport_default,
@@ -297,13 +302,18 @@ def is_valid_keystore(path):
 
 
 def is_key_name_valid(name):
-    # quick check for obvious filesystem problems
-    if ('..' in name) or ('\\' in name) or (not name.startswith('/')):
+    ns_and_name = name.rsplit('/', 1)
+    if len(ns_and_name) != 2:
+        print("The key name needs to start with '/'")
         return False
-    # name must contain characters other than whitespace and '/'
-    if not name.strip('/ '):
+    node_ns = ns_and_name[0] if ns_and_name[0] else '/'
+    node_name = ns_and_name[1]
+
+    try:
+        return (validate_namespace(node_ns) and validate_node_name(node_name))
+    except (InvalidNamespaceException, InvalidNodeNameException) as e:
+        print('{}'.format(e))
         return False
-    return True
 
 
 def create_request_file(path, name):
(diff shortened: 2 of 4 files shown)


jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7294/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3469/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5969/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7113/)

mikaelarguedas:
This was written at the time where node namespace / names conventions were not enforced.
I don't think it's necessarily a good idea /maintainable to extend this custom logic to cover a single new specific case.
Would it be possible to leverage/provide code allowing to validate node namespace/name instead? something like https://github.com/ros2/rmw/blob/master/rmw/include/rmw/validate_node_name.h

jacobperron:
Good idea. Considering this is a Python package, what do you think about depending on `rclpy`, which has functions for validating node names and namespaces?

mikaelarguedas:
That sounds good :+1:

What looks like the ideal outcome would be to have these utilities in a separate package and be used by both `rclpy` and `sros2` (or any other tool needing them).
A commandline argument added here allowing to bypass the check could be valuable as well (imagining people writing pure DDS code and leveraging sros2 to generate their artifacts).

As this is more work (and would require to preserve the rclpy API if fix for Dashing) than directly using the rclpy functions, it may make sense to just do that for now and ticket the remaining items for a later date.

jacobperron:
What's wrong with leaving the validation functions in rclpy? It seems to me that if any other Python tool wants to validation ROS names then the Python ROS client library would be the package to rely on.
I'm not sure what we'd call the new package containing the name validation functions.

jacobperron:
Now using rclpy functions for validation.
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7300/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3475/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5975/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7120/)

mikaelarguedas:
> It seems to me that if any other Python tool wants to validate ROS names then the Python ROS client library would be the package to rely on.

That's a good point :+1:
lgtm as is then

PR #124, opened 2019-06-12
Add basic create_keystore test.

emersonknapp:
Also add dependency on cryptography module to parse generated files.

Signed-off-by: Emerson Knapp <eknapp@amazon.com>

Issue #109, opened 2019-04-06
Replace openssl subprocess calls with python cryptography library


ruffsl:
Currently, the sros2 api makes subprocess calls to the systems OpenSSL CLI for generating public/private x509 key pairs and signing for SMIME. This is a bit hacky, as it relies upon the OpenSSL CLI to be consistent across runtime targets, making it quite fragile should the CLI or implicit defaults change. Instead, we should seek to use a proper cryptography library API so we may have finer control over key material provisioning and error handling for the user.

For example, users wishing to use Connext for DDS Security must swap the environment to point to a version of OpenSSL shipped by RTI. As RTI’s OpenSSL installer does not fully configure the CLI, as it does instead for the shared library, nor does the install respect the system config defaults for OpenSSL; this results in warnings produced by the RTI OpenSSL CLI binary that are susiquentl silenced by sros2 use of subprocess, obscuring potential errors or deviating crypto settings.

---

## Background

> `cryptography` is a package which provides cryptographic recipes and primitives to Python developers. Our goal is for it to be your "cryptographic standard library". `cryptography` includes both high level recipes and low level interfaces to common cryptographic algorithms such as symmetric ciphers, message digests, and key derivation functions.  
https://cryptography.io

By far, my favorite python library for this is `cryptography`, which has a pythonic API, excellent documentation, and readily available for most targets. This what I also used back for SROS1 when developing its keyserver:  
https://github.com/ros/ros_comm/blob/f95b4a5de2acb4fb53f0e9a4cff47dcef928eac5/tools/rosgraph/src/rosgraph/key_helper.py

## Progress

While sros2 CLI was still in its early stages, I previously prototyped a more advanced keystore workspace tool called [Keymint](https://github.com/keymint/keymint_tools), again using the python `cryptography` library:

https://github.com/keymint/keymint_keymake/tree/adc38e07ce5f16d6ba4b36294d7d2e8a361153f0/keymint_keymake/pki

This allowed me to abstract a bit more allowing users to finally configure the asymmetric key algorithm/size/format, CA hierarchy, period of validity and expiration, file protection, etc. I’d like to port over most of these features directly into sros2, but would like to just start by swapping out the subprocess calls and establishing `cryptography` as a primary depency.

The only road bump is foresee is that of supporting SMIME signatures when notarizing DDS Security governance and permission documents. As of writing, the `cryptography` library doesn’t yet seem to have a simple API for producing smime signature. This has been an open ticket for a while:

[...]

PR #126, opened 2019-06-13
Remove use of subprocess for creating ca key and cert

emersonknapp:
Depends on https://github.com/ros2/sros2/pull/124

Removes first instance of `run_shell_command` subprocess call, beginning the fix for https://github.com/ros2/sros2/issues/109 

Signed-off-by: Emerson Knapp <eknapp@amazon.com>

### DIFF
--- a/sros2/sros2/api/__init__.py
+++ b/sros2/sros2/api/__init__.py
@@ -331,7 +331,6 @@ def create_keystore(keystore_path):
 
 def is_valid_keystore(path):
     res = os.path.isfile(os.path.join(path, 'ca_conf.cnf'))
-    res &= os.path.isfile(os.path.join(path, 'ecdsaparam'))
     res &= os.path.isfile(os.path.join(path, 'index.txt'))
     res &= os.path.isfile(os.path.join(path, 'ca.key.pem'))
     res &= os.path.isfile(os.path.join(path, 'ca.cert.pem'))
--- a/sros2/sros2/api/__init__.py
+++ b/sros2/sros2/api/__init__.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from collections import namedtuple
+import datetime
 import itertools
 import os
 import platform
@@ -20,6 +21,11 @@ import shutil
 import subprocess
 import sys
 
+from cryptography import x509
+from cryptography.hazmat.backends import default_backend as cryptography_backend
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives import serialization
+from cryptography.hazmat.primitives.asymmetric import ec
 from lxml import etree
 
 from rclpy.exceptions import InvalidNamespaceException
@@ -116,6 +122,25 @@ def check_openssl_version(openssl_executable):
         raise RuntimeError('need openssl 1.0.2 minimum')
 
 
+def write_key(
+    key,
+    key_path,
+    encoding=serialization.Encoding.PEM,
+    format=serialization.PrivateFormat.PKCS8,
+    encryption_algorithm=serialization.NoEncryption()
+):
+    with open(key_path, 'wb') as f:
+        f.write(key.private_bytes(
+            encoding=encoding,
+            format=format,
+            encryption_algorithm=encryption_algorithm))
+
+
+def write_cert(cert, cert_path, encoding=serialization.Encoding.PEM):
+    with open(cert_path, 'wb') as f:
+        f.write(cert.public_bytes(encoding=encoding))
+
+
 def create_ca_conf_file(path):
     with open(path, 'w') as f:
         f.write("""\
@@ -189,12 +214,21 @@ def create_ecdsa_param_file(path):
     run_shell_command('%s ecparam -name prime256v1 > %s' % (openssl_executable, path))
 
 
-def create_ca_key_cert(ecdsa_param_path, ca_conf_path, ca_key_path, ca_cert_path):
-    openssl_executable = find_openssl_executable()
-    check_openssl_version(openssl_executable)
-    run_shell_command(
-        '%s req -nodes -x509 -days 3650 -newkey ec:%s -keyout %s -out %s -config %s' %
-        (openssl_executable, ecdsa_param_path, ca_key_path, ca_cert_path, ca_conf_path))
+def create_ca_key_cert(ca_key_out_path, ca_cert_out_path):
+    private_key = ec.generate_private_key(ec.SECP256R1, cryptography_backend())
+    write_key(private_key, ca_key_out_path)
+
+    common_name = x509.NameAttribute(x509.oid.NameOID.COMMON_NAME, u'sros2testCA')
+    builder = x509.CertificateBuilder()\
+        .issuer_name(x509.Name([common_name])) \
+        .serial_number(x509.random_serial_number()) \
+        .not_valid_before(datetime.datetime.today() - datetime.timedelta(days=1)) \
+        .not_valid_after(datetime.datetime.today() + datetime.timedelta(days=3650)) \
+        .public_key(private_key.public_key()) \
+        .subject_name(x509.Name([common_name])) \
+        .add_extension(x509.BasicConstraints(ca=True, path_length=1), critical=True)
+    cert = builder.sign(private_key, hashes.SHA256(), cryptography_backend())
+    write_cert(cert, ca_cert_out_path)
 
 
 def create_governance_file(path, domain_id):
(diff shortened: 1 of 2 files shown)


sros2/sros2/api/__init__.py
```diff
@@ -189,12 +214,21 @@ def create_ecdsa_param_file(path):
...
-    check_openssl_version(openssl_executable)
-    run_shell_command(
-        '%s req -nodes -x509 -days 3650 -newkey ec:%s -keyout %s -out %s -config %s' %
-        (openssl_executable, ecdsa_param_path, ca_key_path, ca_cert_path, ca_conf_path))
+def create_ca_key_cert(ca_key_out_path, ca_cert_out_path):
+    private_key = ec.generate_private_key(ec.SECP256R1, cryptography_backend())
+    write_key(private_key, ca_key_out_path)
+
+    common_name = x509.NameAttribute(x509.oid.NameOID.COMMON_NAME, u'sros2testCA')
+    builder = x509.CertificateBuilder()\
```
jacobperron:
```suggestion
    builder = x509.CertificateBuilder() \
```

sros2/sros2/api/__init__.py
```diff
@@ -116,6 +122,25 @@ def check_openssl_version(openssl_executable):
         raise RuntimeError('need openssl 1.0.2 minimum')
 
 
+def write_key(
+    key,
+    key_path,
+    encoding=serialization.Encoding.PEM,
```
kyrofa: I suggest making the kwargs keyword-only, it leads to an API that is easier to use and easier to change in a backward-compatible manner:

```python3
def write_key(
    key,
    key_path,
    *,
    encoding=serialization.Encoding.PEM,
    format=serialization.PrivateFormat.PKCS8,
    encryption_algorithm=serialization.NoEncryption()
```

sros2/sros2/api/__init__.py
```diff
@@ -116,6 +122,25 @@ def check_openssl_version(openssl_executable):
...
+    encryption_algorithm=serialization.NoEncryption()
+):
+    with open(key_path, 'wb') as f:
+        f.write(key.private_bytes(
+            encoding=encoding,
+            format=format,
+            encryption_algorithm=encryption_algorithm))
+
+
+def write_cert(cert, cert_path, encoding=serialization.Encoding.PEM):
```
kyrofa: Ditto here, I suggest making `encoding` keyword-only.

sros2/sros2/api/__init__.py
```diff
@@ -189,12 +214,21 @@ def create_ecdsa_param_file(path):
...
 
 
-def create_ca_key_cert(ecdsa_param_path, ca_conf_path, ca_key_path, ca_cert_path):
-    openssl_executable = find_openssl_executable()
-    check_openssl_version(openssl_executable)
-    run_shell_command(
-        '%s req -nodes -x509 -days 3650 -newkey ec:%s -keyout %s -out %s -config %s' %
-        (openssl_executable, ecdsa_param_path, ca_key_path, ca_cert_path, ca_conf_path))
+def create_ca_key_cert(ca_key_out_path, ca_cert_out_path):
+    private_key = ec.generate_private_key(ec.SECP256R1, cryptography_backend())
```
kyrofa: Can you add a comment regarding the use of `SECP256R1`? Why are we using that one?

sros2/sros2/api/__init__.py
```diff
@@ -189,12 +214,21 @@ def create_ecdsa_param_file(path):
...
+
+    common_name = x509.NameAttribute(x509.oid.NameOID.COMMON_NAME, u'sros2testCA')
+    builder = x509.CertificateBuilder()\
+        .issuer_name(x509.Name([common_name])) \
+        .serial_number(x509.random_serial_number()) \
+        .not_valid_before(datetime.datetime.today() - datetime.timedelta(days=1)) \
+        .not_valid_after(datetime.datetime.today() + datetime.timedelta(days=3650)) \
+        .public_key(private_key.public_key()) \
+        .subject_name(x509.Name([common_name])) \
+        .add_extension(x509.BasicConstraints(ca=True, path_length=1), critical=True)
```
kyrofa: Cryptography's tutorials establish a pattern that I prefer over the line continuations:

```python3
x509.CertificateBuilder().subject_name(
    subject
).issuer_name(
    issuer
)
# ...
```

That style seems easier to modify without getting errors for forgetting a backslash. Thoughts?

sros2/sros2/api/__init__.py
```diff
@@ -236,24 +270,18 @@ def create_keystore(keystore_path):
...
     if not os.path.isfile(ca_conf_path):
         print('creating CA file: %s' % ca_conf_path)
         create_ca_conf_file(ca_conf_path)
     else:
         print('found CA conf file, not writing a new one!')
 
-    ecdsa_param_path = os.path.join(keystore_path, 'ecdsaparam')
-    if not os.path.isfile(ecdsa_param_path):
-        print('creating ECDSA param file: %s' % ecdsa_param_path)
-        create_ecdsa_param_file(ecdsa_param_path)
```
kyrofa: Do we no longer need this, then? What purpose was it serving?

sros2/sros2/api/__init__.py
```diff
@@ -189,12 +214,21 @@ def create_ecdsa_param_file(path):
     run_shell_command('%s ecparam -name prime256v1 > %s' % (openssl_executable, path))
```
kyrofa: Can we get rid of this function, now? Or do we use it/that file elsewhere?

sros2/sros2/api/__init__.py
emersonknapp: This file was the "elliptic curve parameters" for creating the cert. Since we're building in Python, we don't need to write out a parameter file to use as an input to generate the private key for the CA cert, we just specify in the call here https://github.com/ros2/sros2/pull/126/files#diff-4db4152194feafaa6ef7a7fd1c647c53R218 - instead of writing to a file here https://github.com/ros2/sros2/pull/126/files#diff-4db4152194feafaa6ef7a7fd1c647c53R214

sros2/sros2/api/__init__.py
emersonknapp: Yeah, looks better that way. I was thinking in C syntax, will change

sros2/sros2/api/__init__.py
emersonknapp: I'm not 100% sure on the choice, it's the internal name for what we are specifying in this call  https://github.com/ros2/sros2/pull/126/files#diff-4db4152194feafaa6ef7a7fd1c647c53R214 with the arg "prime256v1". @ruffsl maybe you can shed more light on the choice of curve?

sros2/sros2/api/__init__.py
emersonknapp: We are still using `create_ecdsa_param_file` in one other place, my thinking is the next pass on this shell-removal process would remove this function and file. (the linter would complain if it was unused)

sros2/sros2/api/__init__.py
kyrofa: Excellent, thank you.

sros2/sros2/api/__init__.py
```diff
@@ -116,6 +122,26 @@ def check_openssl_version(openssl_executable):
         raise RuntimeError('need openssl 1.0.2 minimum')
 
 
+def write_key(
```
kyrofa: This is looking good, I just have one last question: this API is public, no? Are we sure we want to make functions like this public as well? Or should we prepend them with an underscore? If so, we should do that now, otherwise changing it in the future is technically breaking API.

sros2/sros2/api/__init__.py
mikaelarguedas: > can shed more light on the choice of curve?

The choice of the curve `prime256v1` is specified by the [DDS-Security specification](https://www.omg.org/spec/DDS-SECURITY/1.1/PDF) in section 9.3.1.
> The public key of the CA shall be either a 2048-bit RSA key [44] or else a 256-bit Elliptic Curve Key for the prime256v1 curve [41], also known as the NIST P-256 curve [42].

`SECP256R1` is another alias for that same curve: https://tools.ietf.org/search/rfc4492#appendix-A so this is the right change :+1: 

> Can you add a comment regarding the use of SECP256R1?

:+1: for adding a comment for future code readers

kyrofa: Thank you @mikaelarguedas, that's exactly the ammo required for a good comment!

sros2/sros2/api/__init__.py
emersonknapp: made private

sros2/sros2/api/__init__.py
emersonknapp: comment added

emersonknapp:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7396/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3553/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6070/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7225/)

emersonknapp:
@kyrofa We'll probably have merge conflicts between this and https://github.com/ros2/sros2/pull/129 - let's get that in first since it was ready first :)

emersonknapp:
@jacobperron since #129 has issues, this one first! ;D

PR #127, opened 2019-06-13
Auto generation interface

ryanewel:
Adds a first draft of an interactive cli tool for generating SROS2 policy documents.

Currently this feature is added as a new verb "amend_policy", but since there is a lot of shared code and functionality between this verb and "generate_policy" it may be better to combine the two.


artivis:
Hi @ruffsl,
we (@ryanewel and myself) were wondering if this new functionality shouldn't be merged to the `generate_policy` verb since they are so close in terms of functionality. 
This interactive behavior could become an option of `generate_policy` such as `generate_policy --interactive`. What is your take on this ?

sros2/test/sros2/commands/security/verbs/test_amend_policy_cli.py
```diff
@@ -0,0 +1,66 @@
...
+from shutil import copyfile
+from test import get_policies_path
+
+from ros2cli import cli
+from sros2.policy import load_policy
+from std_msgs.msg import String
+
+
+def test_ament_policy():
+    with tempfile.TemporaryDirectory() as tmpdir:
```
kyrofa: I know you're copying other tests here, but I've recently learned a better way to do this: [pytest's `tmp_path` fixture](https://docs.pytest.org/en/latest/tmpdir.html#the-tmp-path-fixture). Try this:

```python3
def test_amend_policy(tmp_path):
    # ...
    policy_generated_test_file = str(tmp_path.joinpath('dummy_policy.xml'))
    # ...
```

sros2/test/sros2/commands/security/verbs/test_amend_policy_cli.py
```diff
@@ -0,0 +1,66 @@
...
+
+        copyfile(policy_test_file, policy_generated_test_file)
+
+        try:
+            # Create a publisher and subscription
+            node.create_publisher(String, 'denied_topic', 1)
+            node.create_subscription(String, 'parameter_events', lambda msg: None, 1)
+
+            # Generate the policy for the running node
+            with mock.patch.object(builtins, 'input', lambda _: 'Y'):
```
kyrofa: Missing a test for "no"?

sros2/test/sros2/verb/test_amend_policy.py
```diff
@@ -0,0 +1,280 @@
...
+        return False
+    return all(elements_equal(c1, c2) for c1, c2 in zip(e1, e2))
+
+
+def test_add_permission():
+    parser = etree.XMLParser(remove_blank_text=True)
+
+    # No profile for node
+    no_profile_policy_before = """<policy></policy>"""
+    no_profile_policy_after = """<policy>
```
kyrofa: `textwrap.dedent` will make your OCD happier.

sros2/test/sros2/verb/test_amend_policy.py
```diff
@@ -0,0 +1,280 @@
...
+    if e1.tail != e2.tail:
+        return False
+    if e1.attrib != e2.attrib:
+        return False
+    if len(e1) != len(e2):
+        return False
+    return all(elements_equal(c1, c2) for c1, c2 in zip(e1, e2))
+
+
+def test_add_permission():
```
kyrofa: Any chance we can migrate these tests to #128?

sros2/sros2/verb/amend_policy.py
```diff
@@ -0,0 +1,312 @@
...
+from sros2.verb import VerbExtension
+
+POLICY_FILE_NOT_FOUND = 'Package policy file not found'
+POLICY_FILE_NOT_VALID = 'Package policy file not valid'
+
+Event = namedtuple('Event',
+                   ('node_name', 'permission_type', 'rule_type', 'expression'))
+
+
+def get_FQN(node_name, expression):
```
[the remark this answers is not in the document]
artivis: To answer this comment and others similar, we were waiting on knowing if this should be merged with `generate_policy` or not before operating a wide cleanup/refactoring.

sros2/sros2/verb/amend_policy.py
```diff
@@ -0,0 +1,312 @@
...
+
+        self.keep_cached(event)
+
+        print('\n')
+
+    def main(self, *, args):
+        try:
+            self.policy = self.get_policy(args.policy_file_path)
+        except RuntimeError:
+            return POLICY_FILE_NOT_VALID
```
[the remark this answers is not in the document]
artivis: This does return a non-zero upon failure, the pattern was actually copied from other verbs [e.g. here](https://github.com/ros2/ros2cli/blob/master/ros2pkg/ros2pkg/verb/prefix.py#L41)

sros2/test/sros2/commands/security/verbs/test_amend_policy_cli.py
artivis: Looks like it :+1:

sros2/test/sros2/verb/test_amend_policy.py
artivis: Likely doable.
The whole `test_add_permission` ? And only this one?

thomas-moulard:
@ryanewel what are the next steps for this PR?

prajakta-gokhale:
Hello @kyrofa, we're planning to move this PR forward and would like your opinion on whether `amend_policy` qualifies as a separate verb or if it would make more sense to refactor it as an extension to `generate_policy`.

prajakta-gokhale:
@kyrofa could you please take a look at this?

kyrofa:
Wow @prajakta-gokhale, I never saw your ping from October; please accept my humble apologies for not being responsive. In order to share my opinion, my question above stands, can you tell me what you think? I'll quote:

> More generally, I agree that this does seem like an extension of generate_policy. In fact, is generate_policy not virtually exactly this, but with 'Y' hard-coded as the answer to the prompt? Does that mean that this feature is only literally for the user to review the generated policy? I don't mean to belittle this feature, you guys did great work, but do we feel that there is a lot of added value here beyond what users already have by simply reviewing the policy once generate_policy is done with it?

ryanewel:
> More generally, I agree that this does seem like an extension of generate_policy. In fact, is generate_policy not virtually exactly this, but with 'Y' hard-coded as the answer to the prompt? Does that mean that this feature is only literally for the user to review the generated policy? I don't mean to belittle this feature, you guys did great work, but do we feel that there is a lot of added value here beyond what users already have by simply reviewing the policy once generate_policy is done with it?

I think the unique value this feature could provide is letting someone update an existing policy document after changing their ROS application and explicitly highlighting unintended consequences. Using the `generate_policy` verb you could get the same effect by generating a new policy and then diffing it with the existing one. I think this feature would provide a slicker interface, but that's subjective and may not be worth prioritizing.

jacobperron:
I haven't looked closely at the code, but I agree that the new verb seems to overlap a bit with generate_policy. In the spirit of having well-defined tools for specific jobs, I think we could keep the *amend* feature separate from the *generate* feature. Any improvements made in the new verb related to generating policies c an be merged with generate_policy, and the new amend feature can inspect/amend an existing policy file.

---

FYI, I couldn't get the code in the current state to work for me.

With empty ROS graph:

    $ ros2 security amend_policy foo
    ...
    UnboundLocalError: local variable 'filtered_events' referenced before assignment

With a talker running, I answer all the prompts but then get stuck with the repeated message until I ctrl-C:

    ...
    Scanning for events...
    Scanning for events...
    # Error if I shutdown talker first
    ^CFailed to get_subscriber_names_and_types: Node name not found: ns='/', name='talker', at /tmp/binarydeb/ros-eloquent-rmw-fastrtps-shared-cpp-0.8.1/src/rmw_node_info_and_types.cpp:88
    # Otherwise, no policy file is generated (should it be?)

Adding a timeout stops the repeated messages, but I still don't see a policy file created or an existing one modified.

prajakta-gokhale:
We will not be prioritizing working on this feature, for now this PR can be closed.

PR #128, opened 2019-06-13
Move security policy building into API module 

emersonknapp:
Move security policy building so that it can be more easily reused outside the CLI verb. This specifically enables writing a better `create_permission` test, so we don't have to call the CLI in that test.

### DIFF
--- a/sros2/sros2/api/__init__.py
+++ b/sros2/sros2/api/__init__.py
@@ -33,6 +33,7 @@ from sros2.policy import (
     get_transport_schema,
     get_transport_template,
     load_policy,
+    POLICY_VERSION,
 )
 
 HIDDEN_NODE_PREFIX = '_'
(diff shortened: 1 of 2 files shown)


sros2/sros2/api/__init__.py
```diff
@@ -551,3 +552,55 @@ def generate_artifacts(keystore_path=None, identity_names=[], policy_files=[]):
             create_permissions_from_policy_element(
                 keystore_path, identity_name, policy_element)
     return True
+
+
+class PolicyBuilder(object):
+    """Load and modify an XML security policy."""
+    def __init__(self, policy_file_path):
```
kyrofa: Maybe `def __init__(self, *, policy_file_path=None)` makes more sense, given that it's pretty optional?

sros2/sros2/api/__init__.py
```diff
@@ -551,3 +552,55 @@ def generate_artifacts(keystore_path=None, identity_names=[], policy_files=[]):
             create_permissions_from_policy_element(
                 keystore_path, identity_name, policy_element)
     return True
+
+
+class PolicyBuilder(object):
```
kyrofa: No need to inherit from object in Python3.

mjcarroll:
@emersonknapp You'll probably need to take another look at this to resolve conflicts and address feedback.

emersonknapp:
Yeah, thanks for the bump, this fell by the wayside last week - will take a look

kyrofa:
I chatted with @emersonknapp and he gave me the okay to take this over. I've done that and built upon it, refactoring it a bit more and adding a ton of tests, over in #142. Okay to close this?

emersonknapp:
Yeah sounds good

PR #132, opened 2019-06-20
create_key test: add missing tests

kyrofa:
In preparation for future work, this PR adds a few tests that are missing from `create_key`, specifically tests that verify the request.cnf, permissions.p7s, governance.p7s, and ecdsaparam.

Also rework test module to use module fixtures and multiple smaller test cases instead of one large one. This makes it easier to add more, and makes test failures more meaningful.

cc @jacobperron, @emersonknapp

### DIFF
--- a/sros2/test/sros2/commands/security/verbs/test_create_key.py
+++ b/sros2/test/sros2/commands/security/verbs/test_create_key.py
@@ -102,8 +102,6 @@ def test_cert_pem(node_keys_dir):
 
 
 def test_ecdsaparam(node_keys_dir):
-    # Would really like to verify the signature, but ffi just can't use
-    # that part of the OpenSSL API
     with open(os.path.join(node_keys_dir, 'ecdsaparam')) as f:
         assert f.read() == textwrap.dedent("""\
             -----BEGIN EC PARAMETERS-----
--- a/sros2/test/sros2/commands/security/verbs/test_create_key.py
+++ b/sros2/test/sros2/commands/security/verbs/test_create_key.py
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import configparser
 import os
-import tempfile
+import textwrap
 from xml.etree import ElementTree
 
 import cryptography
@@ -22,9 +23,27 @@ from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import ec
 
-from ros2cli import cli
+import pytest
 
+from ros2cli import cli
 from sros2.api import create_keystore
+
+
+# This fixture will run once for the entire module (as opposed to once per test)
+@pytest.fixture(scope='module')
+def node_keys_dir(tmp_path_factory):
+    keystore_dir = str(tmp_path_factory.mktemp('keystore'))
+
+    # First, create the keystore
+    assert create_keystore(keystore_dir)
+
+    # Now using that keystore, create a keypair along with other files required by DDS
+    assert cli.main(argv=['security', 'create_key', keystore_dir, '/test_node']) == 0
+    node_dir = os.path.join(keystore_dir, 'test_node')
+    assert os.path.isdir(os.path.join(keystore_dir, 'test_node'))
+
+    # Return path to directory containing the node's files
+    return node_dir
 
 
 def load_cert(path):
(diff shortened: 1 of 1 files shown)


sros2/test/sros2/commands/security/verbs/test_create_key.py
```diff
@@ -62,69 +81,106 @@ def verify_signature(cert, signatory):
...
     check_common_name(cert.issuer, u'sros2testCA')
+
+    signatory = load_cert(os.path.join(node_keys_dir, 'identity_ca.cert.pem'))
     assert verify_signature(cert, signatory)
 
 
-def check_permissions_xml(path, signatory):
-    ElementTree.parse(path)
+def test_ecdsaparam(node_keys_dir):
+    # Would really like to verify the signature, but ffi just can't use
```
emersonknapp: Lucky for us we'll be removing this file entirely

kyrofa: Whoops, that comment isn't actually valid for that particular test! I removed it. But yeah, looking forward to getting rid of that one.

sros2/test/sros2/commands/security/verbs/test_create_key.py
```diff
@@ -22,11 +23,29 @@
...
-from ros2cli import cli
+import pytest
 
+from ros2cli import cli
 from sros2.api import create_keystore
 
 
+# This fixture will run once for the entire module (as opposed to once per test)
+@pytest.fixture(scope='module')
+def node_keys_dir(tmp_path_factory):
```
dirk-thomas: How does this function get the argument `tmp_path_factory`?

I am asking because the nightly CI jobs fail the test: http://build.ros2.org/view/Eci/job/Eci__nightly-fastrtps_ubuntu_bionic_amd64/19/testReport/sros2.test.sros2.commands.security.verbs/test_create_key/test_create_key/

The nightly job on ci.ros2.org seems to pass (https://ci.ros2.org/view/nightly/job/nightly_linux_debug/1254/testReport/sros2.test.sros2.commands.security.verbs/test_create_key/test_create_key/). Not sure what is different between those.

kyrofa: Well that's an interesting failure. This is a [built-in pytest fixture](https://docs.pytest.org/en/latest/tmpdir.html). Any chance the version of pytest differs between those two tests?

mikaelarguedas: I faced multiple issues about mismatching versions of pytest recently. My guess is that one of these jobs installs a recent pytest from pip and the other one an older version from apt

dirk-thomas: That is certainly the case: one uses the Debian package from Ubuntu Bionic, the other uses the latest release (currently we pinned 4.x instead of using 5 - but that is a different reason). So the code should work with either of the versions.

kyrofa: Good theory, the one from apt may need to also install `python3-pytest-tempdir`.

dirk-thomas: It will only do so if you declare it explicitly as a test dependency in the package manifest.

mikaelarguedas: > Good theory, the one from apt may need to also install python3-pytest-tempdir (total guess based on the name, but seems plausible).

:+1: that would make sense. To do so it will need to be listed as a test dependency in sros2's package.xml and added to the [rosdep database](https://github.com/ros/rosdistro/blob/master/rosdep/python.yaml) for rosdep to resolve it properly.

kyrofa: Yeah, too bad they're split out like that :man_shrugging: . Let me confirm the theory and I'll take care of it if that happens to be the case.

dirk-thomas: > is there a way to run the failing test on a pull request instead of the one that passes? 

The master branch (which will become Eloquent) doesn't have PR jobs yet. So there is no automated way to do that. If you make a PR we can manually trigger a job for it (http://build.ros2.org/view/Eci/job/Eci__overlay_ubuntu_bionic_amd64/).

> Can we do something to bring those testbeds more closely inline with each other?

Not easily - both systems are designed for very different cases and therefore intentionally using different dependencies (Debian packages vs. latest).

kyrofa: Nope, I was wrong. `python3-pytest-tempdir` is not required. Apparently, despite it not being present in the changelog, `tmp_path_factory` wasn't introduced into v5.0.0 of pytest. I'll propose a fix using an older factory.

For what it's worth, it would help avoid these issues if the "install from source" guide installed Debian packages instead of pip. Otherwise issues like this will continue to crop up, with the CI being the limiting factor. Combined with the fact that these jobs don't seem to be the ones that are typically run on PRs, it means we'll continue to catch these issues only after they land, which isn't ideal.

kyrofa: PR is up at #145.

PR #134, opened 2019-06-20
Follow up 126: factor out hardcoded common name

mikaelarguedas:
Factors out the hardcoded `sros2testCA` into a constant `DEFAULT_COMMON_NAME`.

Once this is merged, the tests checking for the common name can be updated to leverage this new constant as well.

Long term, this should likely be exposed to users so that they can pick the name used instead of this hardcoded one

### DIFF
--- a/sros2/sros2/api/__init__.py
+++ b/sros2/sros2/api/__init__.py
@@ -47,6 +47,8 @@ DOMAIN_ID_ENV = 'ROS_DOMAIN_ID'
 NODE_NAME = namedtuple('NodeName', ('node', 'ns', 'fqn'))
 TOPIC_INFO = namedtuple('Topic', ('fqn', 'type'))
 
+DEFAULT_COMMON_NAME = 'sros2testCA'
+
 
 def get_node_names(*, node, include_hidden_nodes=False):
     node_names_and_namespaces = node.get_node_names_and_namespaces()
@@ -143,8 +145,7 @@ def _write_cert(cert, cert_path, *, encoding=serialization.Encoding.PEM):
 
 
 def create_ca_conf_file(path):
-    with open(path, 'w') as f:
-        f.write("""\
+    conf_string = """\
 [ ca ]
 default_ca = CA_default
 
@@ -197,11 +198,13 @@ string_mask = utf8only
 x509_extensions = root_ca_extensions
 
 [ req_distinguished_name ]
-commonName = sros2testCA
+commonName = {common_name}
 
 [ root_ca_extensions ]
 basicConstraints = CA:true
-""")
+""".format(common_name=DEFAULT_COMMON_NAME)
+    with open(path, 'w') as f:
+        f.write(conf_string)
 
 
 def run_shell_command(cmd, in_path=None):
@@ -220,7 +223,7 @@ def create_ca_key_cert(ca_key_out_path, ca_cert_out_path):
     private_key = ec.generate_private_key(ec.SECP256R1, cryptography_backend())
     _write_key(private_key, ca_key_out_path)
 
-    common_name = x509.NameAttribute(x509.oid.NameOID.COMMON_NAME, u'sros2testCA')
+    common_name = x509.NameAttribute(x509.oid.NameOID.COMMON_NAME, DEFAULT_COMMON_NAME)
     builder = x509.CertificateBuilder(
         ).issuer_name(
             x509.Name([common_name])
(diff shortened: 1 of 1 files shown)


sros2/sros2/api/__init__.py
```diff
@@ -44,14 +44,16 @@
 HIDDEN_NODE_PREFIX = '_'
 DOMAIN_ID_ENV = 'ROS_DOMAIN_ID'
 
-NodeName = namedtuple('NodeName', ('node', 'ns', 'fqn'))
-TopicInfo = namedtuple('Topic', ('fqn', 'type'))
+NODE_NAME = namedtuple('NodeName', ('node', 'ns', 'fqn'))
+TOPIC_INFO = namedtuple('Topic', ('fqn', 'type'))
+
+DEFAULT_COMMON_NAME = 'sros2testCA'
```
kyrofa: I suggest prepending all of these with underscores so they aren't part of the public API, unless we intend for them to be used-- do we?

sros2/sros2/api/__init__.py
```diff
@@ -143,8 +145,7 @@ def _write_cert(cert, cert_path, *, encoding=serialization.Encoding.PEM):
 
 
 def create_ca_conf_file(path):
-    with open(path, 'w') as f:
-        f.write("""\
+    conf_string = """\
```
kyrofa: `textwrap.dedent` might be easier on the eyes, here.

kyrofa:
> Once this is merged, the tests checking for the common name can be updated to leverage this new constant as well.

Is there a reason not to do that here?

sros2/sros2/api/__init__.py
```diff
@@ -44,14 +44,16 @@
 HIDDEN_NODE_PREFIX = '_'
 DOMAIN_ID_ENV = 'ROS_DOMAIN_ID'
 
-NodeName = namedtuple('NodeName', ('node', 'ns', 'fqn'))
-TopicInfo = namedtuple('Topic', ('fqn', 'type'))
+NODE_NAME = namedtuple('NodeName', ('node', 'ns', 'fqn'))
```
emersonknapp: Why are you formatting this typename like a constant?

mikaelarguedas: My bad, there's no reason to do so :+1:, I'll just move the new constant above these types declaration to keep the constants grouped together

sros2/sros2/api/__init__.py
mikaelarguedas: Yeah currently most (if not all) of the API in this package is public.
It deserves a refactor to split out part into different modules and a full API review of sros2, from the top of my head it's unclear which of these could be useful by other modules for reuse.
So it may be more appropriate to address in a later API overhaul PR spanning the entire repo.

sros2/sros2/api/__init__.py
mikaelarguedas: Yeah that occurred to me and I figured I'd keep the diff minimal to ease review

As this is already touching the conf file string, there's not much value in not making it look nicer right away so I'll add it to this PR :+1:

mikaelarguedas:
> Is there a reason not to do that here?

There were (still are?) ongoing PRs adding tests or touching tests using of the hard-coded common name when I wrote these commits. So I'd figure we'll let things settle a bit to then do a full search replace.

kyrofa:
> So I'd figure we'll let things settle a bit to then do a full search replace.

Fair point, as long as we don't forget! :+1:

mikaelarguedas:
> Fair point, as long as we don't forget! +1

Opened #137 so that we don't :+1:

sros2/sros2/api/__init__.py
kyrofa: Yeah no reason to modify the others until an overhaul PR, but I still think it's a good idea not to add even more public API unless we're sure it should be public. `DEFAULT_COMMON_NAME` probably doesn't need to be, would you agree?

sros2/sros2/api/__init__.py
kyrofa: Ah, beautiful, thank you!

sros2/sros2/api/__init__.py
mikaelarguedas: Fine by me: https://github.com/ros2/sros2/pull/134/commits/5422fdbbcebe5f834ab0e3b2c4057b01e725db67

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7414/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3567/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6083/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7239/)

PR #129, opened 2019-06-14
Obtain S/MIME signature using cryptography library

kyrofa:
The governance and permission files must be S/MIME signed. The biggest obstacle to adopting the cryptography library is that it doesn't expose an S/MIME API (see https://github.com/pyca/cryptography/issues/1621 for more information).
    
This PR unblocks this and makes progress on #109 by using the OpenSSL API directly (via cryptography's hazard layer) to obtain S/MIME signatures, and prove the functionality out by using it to sign the governance files.

PR #136, opened 2019-06-20
api: migrate permissions S/MIME to cryptography library

kyrofa:
Building upon #129, the S/MIME-signed permissions files are the cause of the final shell out to `openssl` for S/MIME signatures. This PR migrates that logic to use the cryptography library.

### DIFF
--- a/sros2/sros2/api/__init__.py
+++ b/sros2/sros2/api/__init__.py
@@ -265,24 +265,6 @@ def create_governance_file(path, domain_id):
         f.write(etree.tostring(governance_xml, pretty_print=True))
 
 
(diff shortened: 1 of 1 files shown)


kyrofa:
Alright, this one is also ready for review @jacobperron.

kyrofa:
> It looks like the content of this function is now identical to `create_signed_governance_file`, could they be consolidated into a single function `create_smime_signed_file` that we would use to sign both the permissions and governance files?

Of course, done. Note, however, that this breaks API.

mjcarroll:
As far as breaking API, you are targeting `master`, so it's not an issue for Eloquent.

I think it would require some discussion to backport to dashing, if desired.

PR #138, opened 2019-06-24
api: migrate create_key to cryptography library

### LINKED ISSUE #109: Replace openssl subprocess calls with python cryptography library

kyrofa:
This PR resolves #109 by completing the migration from shelling out to `openssl` to the cryptography library.

### DIFF
--- a/sros2/sros2/api/__init__.py
+++ b/sros2/sros2/api/__init__.py
@@ -14,13 +14,9 @@
 
 from collections import namedtuple
 import datetime
-import itertools
 import os
-import platform
 import shutil
-import subprocess
 import sys
-import textwrap
 
 from cryptography import x509
 from cryptography.hazmat.backends import default_backend as cryptography_backend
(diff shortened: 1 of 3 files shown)


emersonknapp:
Woohoo love deleting code - this also closes https://github.com/ros2/sros2/issues/109? Or is there still followup needed

kyrofa:
Yessir, the commit message should make that happen. This still needs a refactor pretty badly, but that's outside of #109.

Issue #94, opened 2019-03-08
Request service permissions should be included in generated policies


ruffsl:
## Feature request
Opening this ticket for feedback on the approach before starting implementation

#### Feature description

Improve the `generate_policy` verb to ensure service permissions for request are accounted for.

##### Context:
As mentioned in https://github.com/ros2/sros2/pull/91#pullrequestreview-210472015 , policy files that auto generated by the `generate_policy` verb do not provision sufficient permissions for services, and instead only provision `reply` permissions when determined necessary via quiring the ros2 graph API.

##### Problems it will solve:
- Startime failures: nodes may otherwise crash at runtime when failing to initialize a client due to insufficient privileges.
- Runtime failures: nodes may also fail when remote servers refuse to accept requests from initialized client interfaces

#### Implementation considerations

This derives from the fact that the present ros2 graph API does not provide an interface to query for what services a node is assuming the client role. Although expanding the ros2 graph API could resolve this absence of information, in general a larger issue remains in that client interfaces may be temporally created, and thus harder to observe via a singular snapshot observation of the graph's current state. Alternate methods for modeling access control policies, such as inferring requirements from log event files like in AppArmor, could prove more complete in generating sufficient permissions for a target application.

Any feedback on the approach will be very appreciated :)

jacobperron:
> in general a larger issue remains in that client interfaces may be temporally created, and thus harder to observe via a singular snapshot observation of the graph's current state. 

Does this also apply to subscriptions/publishers? Sounds like it deserves a separate issue for discussion.

On the surface of it, expanding the graph API seems like a reasonable solution. I haven't put too much thought into it, but I can imagine the graph API being used by a long-running process that observes a ROS system over its lifetime and then generates a policy at the end. This would solve the issue of missing any short-lived ROS entities.

ruffsl:
> Does this also apply to subscriptions/publishers? Sounds like it deserves a separate issue for discussion.

Yes, this would also apply to pub/sub just as well; topics used by short lived action clients comes to mind. I'll try and write up a separate issue for it then.

> imagine the graph API being used by a long-running process that observes a ROS system over its lifetime and then generates a policy at the end. This would solve the issue of missing any short-lived ROS entities.

Perhaps, I suppose something akin to the sros2 daemon  to

mjcarroll:
@ruffsl I'm moving this to backlog for now based on our waffle triage process, but if you start an implementation, feel free to move to "in progress"

PR #141, opened 2019-07-03
Add request service permissions in generated policies 

skucheria:
Addresses https://github.com/ros2/sros2/issues/94

Requires https://github.com/ros2/rclpy/pull/383

### DIFF
--- a/sros2/package.xml
+++ b/sros2/package.xml
@@ -21,6 +21,7 @@
   <test_depend>ament_pep257</test_depend>
   <test_depend>python3-pytest</test_depend>
   <test_depend>std_msgs</test_depend>
+  <test_depend>std_srvs</test_depend>
   <test_depend>rcl_interfaces</test_depend>
 
   <export>
--- a/sros2/package.xml
+++ b/sros2/package.xml
@@ -21,6 +21,7 @@
   <test_depend>ament_pep257</test_depend>
   <test_depend>python3-pytest</test_depend>
   <test_depend>std_msgs</test_depend>
+  <test_depend>rcl_interfaces</test_depend>
 
   <export>
     <build_type>ament_python</build_type>
--- a/sros2/package.xml
+++ b/sros2/package.xml
@@ -22,8 +22,7 @@
   <test_depend>python3-pytest</test_depend>
   <test_depend>std_msgs</test_depend>
   <test_depend>std_srvs</test_depend>
-  <test_depend>rcl_interfaces</test_depend>
-
+  
   <export>
     <build_type>ament_python</build_type>
   </export>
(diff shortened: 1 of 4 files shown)


kyrofa:
This looks good, and the test is lovely, thank you! However, have you ensured that `create_permission` can actually handle the policy produced by this? It may require some [template changes](https://github.com/ros2/sros2/blob/master/sros2/sros2/policy/templates/dds/permissions.xsl#L142).

jacobperron:
I tried passing a generated policy with a service server and service client running. Passing it to `create_permission` seems to work for me. I believe the [add_two_ints.xml](https://github.com/ros2/sros2/blob/master/sros2/test/policies/add_two_ints.xml) policy file used as part of the unit tests covers this.

I think outside the scope of this PR, but it's probably worth adding integration tests to [test_security](https://github.com/ros2/system_tests/tree/master/test_security) for services (and actions).

sros2/test/sros2/commands/security/verbs/test_generate_policy.py
```diff
@@ -62,6 +64,46 @@ def test_generate_policy():
...
+    with tempfile.TemporaryDirectory() as tmpdir:
+        # Create a test-specific context so that generate_policy can still init
+        context = rclpy.Context()
+        rclpy.init(context=context)
+        node = rclpy.create_node('test_node', context=context)
+
+        try:
+            # Create a server and client
+            node.create_client(GetParameters, 'service_client')
+            node.create_service(GetParameters, 'service_server', lambda request,
```
jacobperron: Since nodes already create a [`GetParameters` service](https://github.com/ros2/rclpy/blob/72f55462e7fb9c0011904da9949557996ba56396/rclpy/rclpy/parameter_service.py#L38-L39) as part of the `ParameterServer`, I suggest using a different service type, maybe from [std_srvs](https://github.com/ros2/common_interfaces/tree/master/std_srvs/srv).

sros2/package.xml
```diff
@@ -21,6 +21,8 @@
   <test_depend>ament_pep257</test_depend>
   <test_depend>python3-pytest</test_depend>
   <test_depend>std_msgs</test_depend>
+  <test_depend>std_srvs</test_depend>
+  <test_depend>rcl_interfaces</test_depend>
```
jacobperron: This test dependency can be removed

skucheria:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7499/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3637/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6145/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7325/)

PR #142, opened 2019-07-04
[WIP] policy: add class representation of XML

kyrofa:
XML is not the easiest thing to parse (or create). However, SROS 2  policies are nicely grouped into objects that lend themselves well to a class representation. This PR adds a set of user-friendly classes that represents an SROS 2 policy and its components that can load and dump themselves from and to XML. It also replaces important strings with enums such that they cannot be mistaken. It then updates `generate_policy` to use this new  system. Finally, it covers the whole thing in tests, which is the primary source of the line count (my apologies)

### DIFF
--- a/sros2/sros2/policy/_expression.py
+++ b/sros2/sros2/policy/_expression.py
@@ -1,0 +1,91 @@
+# Copyright 2019 Canonical Ltd
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import enum
+
+from lxml import etree
+
+from . import PolicyError
+
+
+class UnsupportedExpressionTypeError(PolicyError):
+
+    def __init__(self, type_string: str):
+        super().__init__('Unsupported expression type: {!r}'.format(type_string))
+        self.type = type_string
+
+
+@enum.unique
+class ExpressionType(enum.Enum):
+    # Enum values should map to their XML values
+    TOPIC = 'topic'
+    SERVICE = 'service'
+
+
+class Expression:
+    """Class representation of an expression within a permission of an XML security policy."""
+
+    @classmethod
+    def from_fields(
+            cls, fqn: str, namespace: str, expression_type: ExpressionType,
+            text: str) -> 'Expression':
+        """
+        Create new Expression instance from its fields.
+
+        :param str fqn: The node FQN.
+        :param str namespace: The node namespace.
+        :param ExpressionType expression_type: Type of expression.
+        :param str text: Text of expression.
+
+        :returns: The newly-created expression.
+        :rtype: Expression
+        """
+        expression = etree.Element(expression_type.value)
+        namespace = namespace.rstrip('/')
+
+        if text.startswith(fqn + '/'):
+            text = '~' + text[len(fqn + '/'):]
+        elif text.startswith(namespace + '/'):
+            text = text[len(namespace + '/'):]
+
+        expression.text = text
+
+        return cls(expression)
+
+    def __init__(self, expression: etree.Element) -> None:
+        """
+        Create new Expression instance.
+
+        :param etree.Element expression: XML Element containing the expression.
+        """
+        self._expression = expression
+
+    def get_type(self) -> ExpressionType:
+        """
+        Return the type of the expression.
+
+        :rtype: ExpressionType
+        """
+        try:
+            return ExpressionType(self._expression.tag)
+        except ValueError as e:
+            raise UnsupportedExpressionTypeError(self._expression.tag) from e
+
+    def get_text(self) -> str:
+        """
+        Return the text of the expression.
+
+        :rtype: str
+        """
+        return self._expression.text
--- a/sros2/sros2/policy/_expression.py
+++ b/sros2/sros2/policy/_expression.py
@@ -31,6 +31,7 @@ class ExpressionType(enum.Enum):
     # Enum values should map to their XML values
     TOPIC = 'topic'
     SERVICE = 'service'
+    ACTION = 'action'
 
 
 class Expression:
--- a/sros2/sros2/policy/_permission.py
+++ b/sros2/sros2/policy/_permission.py
@@ -1,0 +1,136 @@
+# Copyright 2019 Canonical Ltd
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import enum
+from typing import List
+
+from lxml import etree
+
+from . import PolicyError
+from ._expression import Expression
+
+
+class UnsupportedPolicyError(PolicyError):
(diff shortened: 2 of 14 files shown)


sros2/test/sros2/commands/security/verbs/test_generate_policy.py
```diff
@@ -51,12 +54,78 @@ def test_generate_policy():
         topics_subscribe_allowed = profile.find(path='topics[@subscribe="ALLOW"]')
         assert topics_subscribe_allowed is not None
 
-        # Verify that the allowed publications include topic_pub and not topic_sub
+        # Verify that the allowed publications include topic_pub and topic_both, but not
+        # topic_sub
         topics = topics_publish_allowed.findall('topic')
         assert len([t for t in topics if t.text == 'topic_pub']) == 1
+        # FIXME: Why the leading slash?
```
kyrofa: This doesn't seem quite right, but this is how the code worked before this change, so I kept it.

sros2/test/sros2/commands/security/verbs/test_generate_policy.py
```diff
@@ -51,12 +54,78 @@ def test_generate_policy():
...
+        assert len([t for t in topics if t.text == '/topic/both']) == 1
         assert len([t for t in topics if t.text == 'topic_sub']) == 0
 
-        # Verify that the allowed subscriptions include topic_sub and not topic_pub
+        # Verify that the allowed subscriptions include topic_sub and topic_both, but not
+        # topic_pub
+        topics = topics_subscribe_allowed.findall('topic')
+        assert len([t for t in topics if t.text == 'topic_sub']) == 1
+        # FIXME: Why the leading slash?
```
kyrofa: This doesn't seem quite right, but this is how the code worked before this change, so I kept it.

sros2/test/sros2/policy/test_expression.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def test_nested_under_namespace():
+    expression = _expression.Expression.from_fields(
+        '/test_node', '/ns', _expression.ExpressionType.TOPIC, '/ns/test_topic')
+    assert expression.get_type() == _expression.ExpressionType.TOPIC
+    assert expression.get_text() == 'test_topic'
+
+    # The special case of namespace being '/' should be properly handled.
+    # FIXME: Is it? I would have expected the leading slash to be stripped.
```
kyrofa: This doesn't seem quite right, but this is how the code worked before this change, so I kept it.

sros2/sros2/policy/_profile.py
```diff
@@ -0,0 +1,133 @@
...
+        return permission
+
+    def get_or_create_permission(
+            self, permission_type: PermissionType, rule_type: PermissionRuleType,
+            rule_qualifier: PermissionRuleQualifier) -> Permission:
+        """
+        Get matching permission or create new one if it doesn't exist.
+
+        :param PermissionType permission_type: The type of permission.
+        :param PermissionRuleType rule_type: The type of rule.
```
ruffsl: What should be done if we need to create a permission with multiple `PermissionRuleType`s?

sros2/sros2/verb/generate_policy.py
```diff
@@ -126,24 +68,68 @@ def main(self, *, args):
...
             for node_name in node_names:
-                profile = self.get_profile(policy, node_name)
+                profile = policy.get_or_create_profile(node_name.node, node_name.ns)
+
                 subscribe_topics = get_subscriber_info(node=node, node_name=node_name)
-                if subscribe_topics:
-                    self.add_permission(
-                        profile, 'topic', 'subscribe', 'ALLOW', subscribe_topics, node_name)
+                for topic in subscribe_topics:
+                    _add_subscribe_topic_expression(profile, node_name, topic)
```
ruffsl: I guess `ALLOW` is made implicit here?
Should we make it a default argument to the function signature?
Or never mind as these private function are outside the larger /profile/ api?

sros2/sros2/verb/generate_policy.py
```diff
@@ -126,24 +68,68 @@ def main(self, *, args):
...
-                        profile, 'service', 'reply', 'ALLOW', reply_services, node_name)
+                for service in reply_services:
+                    _add_reply_service_expression(profile, node_name, service)
+
                 request_services = get_client_info(node=node, node_name=node_name)
-                if request_services:
-                    self.add_permission(
-                        profile, 'service', 'request', 'ALLOW', request_services, node_name)
+                for service in request_services:
+                    _add_request_service_expression(profile, node_name, service)
```
ruffsl: Perhaps out of scope for this PR, but can we infer actions from rosgraph yet?
We could add them here when ready.

sros2/sros2/policy/_expression.py
```diff
@@ -0,0 +1,91 @@
...
+        :returns: The newly-created expression.
+        :rtype: Expression
+        """
+        expression = etree.Element(expression_type.value)
+        namespace = namespace.rstrip('/')
+
+        if text.startswith(fqn + '/'):
+            text = '~' + text[len(fqn + '/'):]
+        elif text.startswith(namespace + '/'):
+            text = text[len(namespace + '/'):]
```
ruffsl: This is a lot cleaner that what I had! :+1:

sros2/test/sros2/policy/test_permission.py
```diff
@@ -0,0 +1,139 @@
...
+
+def test_service_reply_allow_from_fields():
+    _assert_service_reply(_permission.Permission.from_fields(
+        _permission.PermissionType.SERVICE, _permission.PermissionRuleType.REPLY,
+        _permission.PermissionRuleQualifier.ALLOW))
+
+
+def test_unsupported_permission():
+    permission_xml = etree.fromstring(textwrap.dedent("""\
+        <topics publish="ALLOW" subscribe="ALLOW">
```
ruffsl: Why is this unsupported?
These attributes are not purely exclusive.

https://github.com/ros2/sros2/blob/1c274b41cde9ab30220b85b1b82d164dae0d23a1/sros2/sros2/policy/schemas/policy.xsd#L41-L42

Although, for sros, at least one should be set.
But I'm not sure that is expressible in XSD 1.0.
I could be wrong, but the current xsd doesn't seem to enforce this.

sros2/sros2/policy/_permission.py
```diff
@@ -0,0 +1,137 @@
...
+
+@enum.unique
+class PermissionType(enum.Enum):
+    # Enum values should map to their XML values
+    TOPIC = 'topics'
+    SERVICE = 'services'
+
+
+@enum.unique
+class PermissionRuleType(enum.Enum):
```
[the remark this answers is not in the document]
kyrofa: Yeah I agree, I couldn't quite come up with an obvious way to do that while keeping the API simple. Let me mull it over a bit.

sros2/sros2/verb/generate_policy.py
kyrofa: Indeed, this is just breaking up logical steps into functions. It's private and pretty specific, so I think it's okay to assume `ALLOW`. If in the future we need to make it more generically useful, that would be super easy.

sros2/sros2/verb/generate_policy.py
kyrofa: I'm uncertain of the status of that work, but I agree it would be great to have. I also agree it should be another PR though, this one is pretty large.

sros2/sros2/policy/_permission.py
```diff
@@ -0,0 +1,137 @@
...
+    def get_rule_type(self) -> PermissionRuleType:
+        """
+        Return the type of the rule.
+
+        :rtype: PermissionRuleType
+        """
+        keys = self._permission.keys()
+        if len(keys) != 1:
+            raise UnsupportedPolicyError(
+                'Expected a single attribute to determine rule type, got {!r}'.format(keys))
```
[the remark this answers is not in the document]
kyrofa: Yeah, looks like you [saw the test that covers this case](https://github.com/ros2/sros2/pull/142#discussion_r302331698), so we can discuss over there.

sros2/test/sros2/policy/test_permission.py
kyrofa: Mostly because the existing code didn't write XML that had them both, but I realize now that we definitely need to be able to at least _parse_ XML that does, even if we don't write them. I'll fix it.

kyrofa:
Sorry for the delay here, still working through a bit of a redesign.

kyrofa:
This needs more work, no need to keep this open until I'm able to complete it.

PR #145, opened 2019-07-17
tests: use older pytest

kyrofa:
ROS 2 tests are limited by the version of packages available in Ubuntu Bionic, not the ones available in pip. As a result, we need to make sure we only take advantage of the features as of pytest v3.3.2. A side effect of this is that we must use `tmpdir_factory` as opposed to `tmp_path_factory`, which wasn't introduced until v4.3.0.

### DIFF
--- a/sros2/test/sros2/commands/security/verbs/test_create_key.py
+++ b/sros2/test/sros2/commands/security/verbs/test_create_key.py
@@ -34,8 +34,8 @@ from sros2.policy import get_transport_schema
 
 # This fixture will run once for the entire module (as opposed to once per test)
 @pytest.fixture(scope='module')
-def node_keys_dir(tmp_path_factory):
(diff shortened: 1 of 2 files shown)


kyrofa:
Can someone with power please manually trigger a CI job using http://build.ros2.org/view/Eci/job/Eci__overlay_ubuntu_bionic_amd64/ ?

dirk-thomas:
Before: [Build Status](http://build.ros2.org/job/Eci__overlay_ubuntu_bionic_amd64/12/)
After: [Build Status](http://build.ros2.org/job/Eci__overlay_ubuntu_bionic_amd64/13/)

dirk-thomas:
And the usual CI (using the `pip` version of `pytest`): [Build Status](https://ci.ros2.org/job/ci_linux/7600/)

dirk-thomas:
Thanks for the quick patch.

PR #146, opened 2019-07-31
add profile for lifecycle nodes

mikaelarguedas:

What would be a good place for storing common policy profiles in the future? Should they be in the sros2 package and installed by it? or would a separate package be more appropriate?

cc @ruffsl @kyrofa

### DIFF
--- a/sros2/test/policies/common/lifecycle_node.xml
+++ b/sros2/test/policies/common/lifecycle_node.xml
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<profile xmlns:xi="http://www.w3.org/2003/XInclude">
+  <xi:include href="node.xml"
+    xpointer="xpointer(/profile/*)"/>
+  <services reply="ALLOW">
+    <service>~change_state</service>
+    <service>~get_available_states</service>
+    <service>~get_available_transitions</service>
+    <service>~get_state</service>
+    <service>~get_transition_graph</service>
+  </services>
+  <topics publish="ALLOW">
+    <topic>~transition_event</topic>
+  </topics>
+</profile>
--- a/sros2/test/policies/common/lifecycle_node.xml
+++ b/sros2/test/policies/common/lifecycle_node.xml
@@ -4,13 +4,13 @@
   <xi:include href="node.xml"
     xpointer="xpointer(/profile/*)"/>
   <services reply="ALLOW">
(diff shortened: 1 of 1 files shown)


ruffsl:
With development of https://github.com/ros2/sros2/pull/147, it would make sense create like `sros2_common_`something. I'm not sure what to call it as it could be used to hold policies, profiles, or profiles snippets. But the gist of housing common ones in a separate package seems good.

kyrofa:
I'm having trouble wrapping my head around how much this turns into. I feel like it could be a lot, though, so I tend to agree that it something be held in a dedicated place. `sros2_common_assets`?

mikaelarguedas:
> With development of #147, it would make sense create like sros2_common_something. I'm not sure what to call it as it could be used to hold policies, profiles, or profiles snippets

Hosting in a separate package rather than in sros2 itself seems sensible :+1:

> I'm having trouble wrapping my head around how much this turns into. I feel like it could be a lot, though, so I tend to agree that it should be held in a dedicated place. sros2_common_assets?

`sros2_common_assets` sounds good to me. I would also be fine with `sros2_common_policies` as I see profiles and profiles snippets as a subset of policies (similar to how docker snippets are hosted in the [docker_templates repository](https://github.com/osrf/docker_templates/tree/master/docker_templates/templates/snippet))

---

I don't have a strong preference, we can either merge this now or wait for #147 to land and then target this at the new package.

mikaelarguedas:
@mjcarroll @jacobperron is it possible to merge this ? I'm happy to iterate once #147 is picked up

jacobperron:
@mikaelarguedas Just to clarify, what is the intended purpose for the new profile? It appears in the `test` directory and is not used by any tests nor is it installed.

mikaelarguedas:
The goal is to have all the common ROS primitive policies stored at the same place. This way they can be easily copied and reused. But they won't be installed until the work to actually install and register policies is complete (see #147).

The fact that the lifecycle ones are not on any "official" branch makes it harder to track changes such as https://github.com/ros2/sros2/commit/b10a7e95a63692ab9632532dd8b8989e0be58fef

The value of such policies is to simplify policy files such as https://github.com/ros-swg/turtlebot3_demo/commit/2d394cbb90d1aab56768bdc1166fbc2f362117d3#diff-d989d1536b8a5fbbdf2e4f029676fc6b

PR #147, opened 2019-07-31
[WIP] Register common policies with the ament index

mjcarroll:


sros2_common_policies/policies/node.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<profile xmlns:xi="http://www.w3.org/2003/XInclude">
+  <sros2_policy package="sros2_common_policies" policy="node/logging.xml" />
```
ruffsl: I think I'd prefer that assets inside the same package keep with using the conventional XML xinclude for importation, so that a preprocessor to recursively expand  `<sros2_policy>` tags would not be a requirement in reusing assets in the sros2 common package.

PR #152, opened 2019-08-04
topics starting with tilde need a slash right after

mikaelarguedas:
from http://design.ros2.org/articles/topic_and_service_names.html:
> must separate a tilde (~) from the rest of the name with a forward slash (/), i.e. ~/foo not ~foo

Note that there is currently no enforcement that the node or topic names provided in the policy files are valid.
In the future we could consider checking the validity of the node and topic names using the validation functions provided by ROS.
If we do so we may want to introduce a way to bypass the check, and/or better, a separate tag for pure transport topics (to which ROS conventions would not be applied)

### DIFF
--- a/sros2/sros2/policy/templates/dds/permissions.xsl
+++ b/sros2/sros2/policy/templates/dds/permissions.xsl
@@ -235,7 +235,7 @@
         </xsl:call-template>
       </xsl:variable>
       <xsl:variable name="_name" select="substring($name, 2)"/>
-      <xsl:value-of select="concat($_ns, $node, '/', $_name)"/>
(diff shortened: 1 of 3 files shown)


jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7847/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3914/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6403/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7693/)

Issue #112, opened 2019-05-01
Auto generation interface for Access Control Profiles


ruffsl:
Formulating access control profiles for large or complex ROS applications that satisfy [*principles of least privilege*](https://en.wikipedia.org/wiki/Principle_of_least_privilege) is at a present manual endeavor for robotics. Given the tedium of such tasks and the rigor it demands to be fully effective, instances of human error become both more probable and problematic; as accidental over provisioning of permissions presents issues in system security, while under provisioning may similarly effect system stability.

To ease the development of complete and correct access control, automated tooling for profile generation must be developed. Such tooling could consume system measurements such as security event logs or discovery traffic to assist in accurately profiling all permissions required. Subsequently, the acquired permission model could then be exported into profile formats used in SROS2 keystore tooling; e.g. a CLI exposing an interactive session for amending work-in-progress profiles by prompting the user about discrepancies between the existing xml policy file vs. the acquired model.

---

## Background

> SROS1 provides a varying degree of run-time modes, including audit, enforce and complain, again borrowing feature designs from AppArmor. This provides developers a method to auto generate, or amend profiles through granular logging of access events and violation attempts.
>
> “SROS1: Using and Developing Secure ROS1 Systems”
> https://doi.org/10.1007/978-3-319-91590-6_11

Auto generation of profiles was used in SROS1 to simplify the bootstrapping process for enabling fine-grained API access control over ROS1. As ROS1 was centralized with respect to the master node, roscore was extended to support several run-time modes; specifically an audit mode that could amend any work-in-progress profile to include missing permission for API accesses observed during training time. This profile was then used in conjunction with a keyserver process the provisioned the certificates and tokens in keystore accordingly.

> Security of robotics systems, as well as of the related middleware infrastructures, is a critical issue for industrial and domestic IoT, and it needs to be continuously assessed throughout the whole development lifecycle... In this work, we introduce a framework for procedural provisioning access control policies for robotic software, as well as for verifying the compliance of generated transport artifacts and decision point implementations.
> 
> “Procedurally Provisioned Access Control for Robotic Systems”
> https://doi.org/10.1109/IROS.2018.8594462

In later work, some of this bootstrapping was reproduced for ROS2 to assist in the automation for the verification and testing of security artifacts used for secure transports. Specifically, given decentralized networking of ROS2 with DDS, observed discovery traffic during training time was substituted to train the model of necessary API permission.

[...]

PR #153, opened 2019-08-06
Pipeline for policy generation from DDS discovery data

ruffsl:
Related: https://github.com/ros2/sros2/issues/112


sros2/sros2/policy/templates/dds/policy.xsl
```diff
@@ -0,0 +1,494 @@
...
+    <mapping type="reply"    prefix="rr"  suffix="/_action/get_resultReply"/>
+    <mapping type="reply"    prefix="rr"  suffix="/_action/send_goalReply"/>
+    <mapping type="request"  prefix="rq"  suffix="/_action/cancel_goalRequest"/>
+    <mapping type="request"  prefix="rq"  suffix="/_action/get_resultRequest"/>
+    <mapping type="request"  prefix="rq"  suffix="/_action/send_goalRequest"/>
+    <mapping type="feedback" prefix="rt"  suffix="/_action/feedback"/>
+    <mapping type="status"   prefix="rt"  suffix="/_action/status"/>
+  </mappings>
+  <permissions>
+    <permission kind="publish"   type="feedback" rule="excute"/>
```
mikaelarguedas:
```suggestion
    <permission kind="publish"   type="feedback" rule="execute"/>
```

sros2/scripts/dds_sql_to_sros2_policy.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+# with pd.option_context('display.max_rows', None, 'display.max_columns', None):  # more options can be specified also
+#     display(df)
+
+
+# In[ ]:
+
+
+polciy = etree.Element('polciy')
```
mikaelarguedas:
```suggestion
policy = etree.Element('policy')
```

mikaelarguedas: Same multiple times below

sros2/scripts/dds_sql_to_sros2_policy.py
```diff
@@ -0,0 +1,172 @@
...
+for namespace in df.index.get_level_values('namespace').unique():
+    for name in df.index.get_level_values('name').unique():
+        profile = etree.SubElement(profiles, 'profile')
+        profile.set("ns", namespace)
+        profile.set("node", name)
+        _df = df['dds_topic'].loc[namespace, name, :]
+        for mode in _df.index.get_level_values('mode').unique():
+            topics = etree.SubElement(profile, 'dds_topics')
+            topics.set(mode, "ALLOW")
+            for dds_topic in df['dds_topic'].loc[namespace, name, :]:
```
mikaelarguedas: I believe this should check for `mode` to not add all topics corresponding to this namespace + name

mikaelarguedas (commented):
Some things that came up when spot comparing the generated files with the ones from https://github.com/mikaelarguedas/tb3_demo/blob/857e1c62ba06476cda4ba359fdf59d31ba488f91/policies/sqlite_playground/extract_policy_from_discovery_db.py

ruffsl:
How's it now? See example output:
https://pastebin.com/JhCvvYVh

sros2/scripts/dds_sql_to_sros2_policy.py
```diff
@@ -0,0 +1,143 @@
...
+
+    db_uri = 'file:{}?mode=ro'.format(args.input_db)
+    with sqlite3.connect(db_uri, uri=True) as db:
+        df = db_to_df(db)
+    
+    dds_policy = df_to_dds_policy(df)
+    sros2_policy = dds_policy_to_sros2_policy(dds_policy)
+
+    if args.output_policy is not None:
+        with open(args.policy_file, 'wb') as f:
```
mikaelarguedas:
```suggestion
        with open(args.output_policy, 'wb') as f:
```

sros2/scripts/dds_sql_to_sros2_policy.py
```diff
@@ -0,0 +1,143 @@
+#! /usr/bin/env python3
+
+import argparse
+import os
+import pandas as pd
```
mikaelarguedas: Note for later: needed to install this dep by hand, we should list it in the `package.xml` ([python3-pandas](https://github.com/ros/rosdistro/blob/a63054565aa50062f25d4fc558e5555b989587ef/rosdep/python.yaml#L5088)) when this is ready for prime time

mikaelarguedas (commented):
> How's it now?

Better! less unnecessary permissions and it validates the schema!

Few surprising things in the result file:

- empty profiles:
Profiles are generated for namespaces that don't belong to them and this results in a lot of empty profiles.
```
    <profile node="gazebo" ns="/global_costmap"/>
    <profile node="gazebo" ns="/local_costmap"/>
```

- There are some topics listed in the database that don't appear in the generated file, for example `tf` doesnt appear for any node in the generated policies o_O

- `clock` appears sometimes absolute and sometimes relative. I believe it makes sense as when the namespace is just `/` the script has no way to know which one of the two it is. Something to keep in mind when adding logic to collapse permissions to higher level common objects

ruffsl:
> * empty profiles:
>   Profiles are generated for namespaces that don't belong to them and this results in a lot of empty profiles.

Empty profiles were unintendedly added with an erroneous combinatorial across nodes and namespaces: address in 368f4c5 . I've also added a check for when the discovery user_data from a foreign QoS configuration: see commit message in 3533ec2

> * There are some topics listed in the database that don't appear in the generated file, for example `tf` doesnt appear for any node in the generated policies o_O

The template can be debugged better commenting back the secondary transforms:

https://github.com/ros2/sros2/blob/1854b7ec50c47c0be3c2dce384b0b053c10d6e88/sros2/sros2/policy/templates/dds/demangle.xsl#L204-L215

Diffing the consecutive chained output from different iterations of commiting, I found some copy and paste typos in the compression. See 1854b7e . I'd like to make this a little more DRY, but it's tricky when only using XSL v1.0. Let me know if you see a better expression for things. 

> * `clock` appears sometimes absolute and sometimes relative. I believe it makes sense as when the namespace is just `/` the script has no way to know which one of the two it is. Something to keep in mind when adding logic to collapse permissions to higher level common objects

Yea, I thought about this, like adding a lookup for common root level topics in the xst for such cases, but it's probably better instead to ensure the added logic to collapse permissions to higher level common objects is robust enough to handle both appearances, perhaps by expanding the fqn topic namespace using rcl logic when comparing candidate substitutions?

sros2/scripts/dds_sql_to_sros2_policy.py
ruffsl: See: 0b27a81

ruffsl:
You can diff from last pastebin using the fix output here: https://pastebin.com/WYvxNTbg

sros2/scripts/dds_sql_to_sros2_policy.py
ruffsl: 297b049

mikaelarguedas:
Deduplicating the topics in the raw intermediate representation now generate the exact same topic list as https://github.com/mikaelarguedas/tb3_demo/blob/857e1c62ba06476cda4ba359fdf59d31ba488f91/policies/sqlite_playground/policy_dump_from_sqlite.xml#L1094 :tada: 

```diff
diff --git a/sros2/scripts/dds_sql_to_sros2_policy.py b/sros2/scripts/dds_sql_to_sros2_policy.py
index 6c602cd..801a919 100755
--- a/sros2/scripts/dds_sql_to_sros2_policy.py
+++ b/sros2/scripts/dds_sql_to_sros2_policy.py
@@ -91,9 +91,12 @@ def df_to_dds_policy(df):
                 if not pd.isna(mode):
                     topics = etree.SubElement(profile, 'raws')
                     topics.set(mode, "ALLOW")
+                    raw_topics = []
                     for dds_topic in df['dds_topic'].loc[namespace, name, mode]:
-                        topic = etree.SubElement(topics, 'raw')
-                        topic.text = dds_topic
+                        if dds_topic not in raw_topics:
+                            raw_topics.append(dds_topic)
+                            topic = etree.SubElement(topics, 'raw')
+                            topic.text = dds_topic
     return dds_policy
```

Comparing the second stage is a bit trickier but spot checking a few nodes everything seems to work as expected :+1:

>  I'd like to make this a little more DRY, but it's tricky when only using XSL v1.0. Let me know if you see a better expression for things.

Yeah that would be good indeed, but like you said it looks tricky to do with XSL 1.0. We can do more investigation after we get everything running well.

>  like adding a lookup for common root level topics in the xst for such cases, but it's probably better instead to ensure the added logic to collapse permissions to higher level common objects is robust enough to handle both appearances, perhaps by expanding the fqn topic namespace using rcl logic when comparing candidate substitutions?

Yep that's the approach I had in mind (using FQNs) before comparing to common assets.
Not sure if rcl logic needs to be leveraged here as the concatenation of namespace and names already happens in the xsl in some places.

ruffsl:
@mikaelarguedas and I chatted offline about adding a `<raw>` type element to the policy schema. Though I think we'll hold off on that for now until we have a more concrete use case for it. So I've just relegated that schema check to its own transport folder for now. a67dbc7

PR #154, opened 2019-08-09
Add mypy tests to check static typing

kyrofa:
This uses the newly-added ament_mypy package.

### DIFF
--- a/sros2/setup.py
+++ b/sros2/setup.py
@@ -12,11 +12,11 @@ def package_files(directory):
     return paths
 
 
-extra_files = []
-extra_files.extend(package_files('sros2/policy/defaults'))
-extra_files.extend(package_files('sros2/policy/schemas'))
-extra_files.extend(package_files('sros2/policy/templates'))
-
+extra_files = (
+    package_files('sros2/policy/defaults') +
+    package_files('sros2/policy/schemas') +
+    package_files('sros2/policy/templates')
+)
 
 setup(
     name='sros2',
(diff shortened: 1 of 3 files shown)


sros2/test/test_mypy.py
```diff
@@ -0,0 +1,22 @@
...
+# limitations under the License.
+
+from ament_mypy.main import main
+import pytest
+
+
+@pytest.mark.mypy
+@pytest.mark.linter
+def test_mypy():
+    assert main(argv=[]) == 0, 'Found errors'
```
ruffsl: How does this innovation work?
Does it just test all source .py files in the parent package the test file is located under?

kyrofa: Yes, at this point it will be the .py files in this package only. We went back and forth on the PR setting the `MYPYPATH` in different ways, but ended up shipping the simplest thing that worked while we continue discussing how that should be done.

sros2/setup.py
```diff
@@ -12,11 +12,11 @@ def package_files(directory):
     return paths
 
 
-extra_files = []
-extra_files.extend(package_files('sros2/policy/defaults'))
-extra_files.extend(package_files('sros2/policy/schemas'))
-extra_files.extend(package_files('sros2/policy/templates'))
-
+extra_files = (
```
jacobperron: Is this change relevant, and if so how?

kyrofa: It allows mypy to determine the type without us needing to add it manually.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7937/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3971/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6462/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7786/)

jacobperron:
Thanks for the addition!

PR #159, opened 2019-09-15
install package manifest

ruffsl:
For osrf/docker_images#317.

### DIFF
--- a/sros2/setup.py
+++ b/sros2/setup.py
@@ -18,10 +18,15 @@ extra_files = (
     package_files('sros2/policy/templates')
 )
 
+package_name = 'sros2'
+
 setup(
-    name='sros2',
+    name=package_name,
     version='0.6.2',
     packages=find_packages(exclude=['test']),
+    data_files=[
+        ('share/' + package_name, ['package.xml']),
+    ],
     install_requires=['setuptools'],
     zip_safe=True,
     author='Morgan Quigley',


sros2/setup.py
```diff
@@ -11,17 +11,21 @@ def package_files(directory):
             paths.append(os.path.join('..', path, filename))
     return paths
 
-
```
jacobperron: The linter is complaining about this change.

ruffsl: @jacobperron , I've reverted this line.

jacobperron:
Sorry for the delay.

@ruffsl Please do a squash-merge or rebase-merge next time to avoid an extra merge commit.

ruffsl:
> Please do a squash-merge or rebase-merge next time to avoid an extra merge commit.

Can you set that as the default option for the repo in the github repo settings?
It's not apparent which megre style is prefered for every github repo otherwise.

jacobperron:
I've updated the settings to disallow merge commits. I'm not sure if there's a reason we don't do this across the org, so I'll only touch the settings for this repo for now.

mjcarroll:
I know that in some cases we don't do squash-merge to preserve the history.  The case that comes to mind is if there is a commit that represents a copy from somewhere else, and then additional commits for modifications to that copy.

It should also default to the last one that you used on that repository.

PR #161, opened 2019-09-27
Install local cache of XML schemas on 'dashing'

pbaughman:
This is the 'dashing' version of [this PR](https://github.com/ros2/sros2/pull/158)

### DIFF
--- a/sros2/setup.py
+++ b/sros2/setup.py
@@ -1,3 +1,4 @@
+import glob
 import os
 
 from setuptools import find_packages
@@ -28,6 +29,7 @@ setup(
         ('share/' + package_name, ['package.xml']),
         ('share/ament_index/resource_index/packages',
             ['resource/' + package_name]),
+        ('share/sros2/xml_cache', glob.glob('xml_cache/**')),
     ],
     install_requires=['setuptools'],
     zip_safe=True,
--- a/sros2/package.xml
+++ b/sros2/package.xml
@@ -14,6 +14,7 @@
   <depend>ros2cli</depend>
 
   <exec_depend>openssl</exec_depend>
+  <exec_depend>ament_index_python</exec_depend>
   <exec_depend>python3-lxml</exec_depend>
 
   <test_depend>ament_copyright</test_depend>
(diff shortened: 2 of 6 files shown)


pbaughman:
@kyrofa sros2 already has a [runtime dependency on rclpy](https://github.com/ros2/sros2/blob/dashing/sros2/package.xml#L13), which itself has a [runtime dependency on ament_index_python](https://github.com/ros2/rclpy/blob/dashing/rclpy/package.xml#L26), so anybody running sros2 should already have ament_index_python available at runtime.

kyrofa:
> anybody running sros2 should already have ament_index_python available at runtime.

Ah, you covered your bases. I was about to start in on how `apt-get upgrade` won't install this update with the new dependency, but I don't think it'll complain if the dependency is already installed.

sros2/setup.py
```diff
@@ -18,10 +19,18 @@ def package_files(directory):
...
+
 setup(
-    name='sros2',
+    name=package_name,
     version='0.6.2',
     packages=find_packages(exclude=['test']),
+    data_files=[
+        ('share/' + package_name, ['package.xml']),
+        ('share/ament_index/resource_index/packages',
+            ['resource/' + package_name]),
```
kyrofa: These lines are fixing an unrelated issue-- mind extracting them into their own PR? Thanks for the fix, though!

pbaughman: @kyrofa Yeah, I can open a PR for 
9b20326 and then rebase 045699d once it's merged

pbaughman: @kyrofa Done https://github.com/ros2/sros2/pull/162

pbaughman: @kyrofa Ok, this one is rebased now so it just contains one commit relevant to the XML catalog

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8241/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4209/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6723/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8132/)

PR #168, opened 2019-10-28
Add delay after node creation to ensure discovery for other rmw implementations

BMarchi:
For connext, there were two issues:
- https://ci.ros2.org/job/ci_linux/8479/testReport/junit/sros2.test.sros2.commands.security.verbs/test_generate_policy/test_generate_policy_services/
- https://ci.ros2.org/job/ci_linux/8479/testReport/junit/sros2.test.sros2.commands.security.verbs/test_generate_policy/test_generate_policy_topics/

When the `cli` command is run, https://github.com/ros2/sros2/blob/c7ae64e6f734abab79c9c3e9d6f0da3125d33ba2/sros2/test/sros2/commands/security/verbs/test_generate_policy.py#L41-L42

it runs in another thread/context, ending up here:
https://github.com/ros2/sros2/blob/c7ae64e6f734abab79c9c3e9d6f0da3125d33ba2/sros2/sros2/verb/generate_policy.py#L130

The culprit here, is the next portion of code:
https://github.com/ros2/rmw_connext/blob/b1062f92bfbec2e85be3083144a0f6c5a599287d/rmw_connext_shared_cpp/src/node_info_and_types.cpp#L104-L139

`get_discovered_participants` from connext is returning 0 handlers, concluding that connext is not finding the previously created node in python.
A timer is not ideal, but for testing purposes I see no other way to ensure the node's advertising.

### DIFF
--- a/sros2/test/sros2/commands/security/verbs/fixtures/client_service_node.py
+++ b/sros2/test/sros2/commands/security/verbs/fixtures/client_service_node.py
@@ -12,8 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import sys
-
 import rclpy
 from rclpy.node import Node
 
@@ -43,9 +41,6 @@ def main(args=None):
         rclpy.spin(node)
     except KeyboardInterrupt:
         print('node stopped cleanly')
-    except BaseException:
-        print('exception in node:', file=sys.stderr)
-        raise
     finally:
         node.destroy_node()
         rclpy.shutdown()
--- a/sros2/test/sros2/commands/security/verbs/fixtures/client_service_node.py
+++ b/sros2/test/sros2/commands/security/verbs/fixtures/client_service_node.py
@@ -1,0 +1,55 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+
+import rclpy
+from rclpy.node import Node
+
+from std_srvs.srv import Trigger
+
+
+class ClientServiceNode(Node):
+
+    def __init__(self):
+        super().__init__('test_generate_policy_services_node')
+        self.client = self.create_client(Trigger, 'test_generate_policy_services_client')
+        self.service = self.create_service(
+            Trigger, 'test_generate_policy_services_server', lambda request, response: response)
+
+    def destroy_node(self):
+        self.client.destroy()
+        self.service.destroy()
+        super().destroy_node()
+
+
+def main(args=None):
+    rclpy.init(args=args)
+
+    node = ClientServiceNode()
+
+    try:
+        rclpy.spin(node)
+    except KeyboardInterrupt:
+        print('node stopped cleanly')
+    except BaseException:
+        print('exception in node:', file=sys.stderr)
+        raise
+    finally:
+        node.destroy_node()
+        rclpy.shutdown()
+
+
+if __name__ == '__main__':
+    main()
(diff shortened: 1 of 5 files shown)


BMarchi:
I was checking your suggestion @hidmic, but sometimes the test fails, even with my changes. Since we have to make retries and specify longer timeouts for `ros2cli` tests for every rmw_implementation for example, shouldn't we do the same thing here since it's using a `cli` feature?

sros2/test/sros2/commands/security/verbs/test_generate_policy.py
```diff
@@ -36,7 +37,8 @@ def test_generate_policy_topics():
             node.create_publisher(String, 'test_generate_policy_topics_pub', 1)
             node.create_subscription(
                 String, 'test_generate_policy_topics_sub', lambda msg: None, 1)
-
+            # Give time to some rmw implementations to advertise the nodes
+            time.sleep(.500)
             # Generate the policy for the running node
             assert cli.main(
```
jacobperron: Looping here until `cli.main(...)` evaluates to `True` or a timeout occurs should be more robust. I believe this is what @hidmic is suggesting. Same for the other test below.

BMarchi: Yeah, for some reason now the tests do not pass at all even waiting for more than 1.5 seconds. A better thing to do instead of iterating and waiting will be to use `launch_testing`

hidmic: +1

BMarchi:
I changed the test to start using launch_testing instead, take a look whenever you can @jacobperron , @hidmic

BMarchi:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/8751/)
- Arch [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4580/)
- macOS [Build Status](http://ci.ros2.org/job/ci_osx/7153/)
- Windows [Build Status](http://ci.ros2.org/job/ci_windows/8676/)

sros2/test/sros2/commands/security/verbs/test_generate_policy_no_nodes.py
```diff
@@ -0,0 +1,26 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import tempfile
+
+from ros2cli import cli
+
+
+def test_generate_policy_no_nodes(capsys):
```
jacobperron: I'm not sure that this PR resolves the previous test failure on Windows (https://github.com/ros2/sros2/issues/143).
We probably still want to keep the TODO and skip decorator. Note, to reproduce the failure, you should also run tests for packages other than sros2 (e.g. `colcon test --packages-select demo_nodes_cpp && colcon test --packages-select sros2`).

BMarchi:
Re run CI for linux:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/8757/)

Windows nodes are dead

sros2/test/sros2/commands/security/verbs/fixtures/pub_sub_node.py
```diff
@@ -0,0 +1,55 @@
...
+
+    node = PubSubNode()
+
+    try:
+        rclpy.spin(node)
+    except KeyboardInterrupt:
+        print('node stopped cleanly')
+    except BaseException:
+        print('exception in node:', file=sys.stderr)
+        raise
```
hidmic: @BMarchi nit: isn't this `except` clause somewhat redundant?

sros2/test/sros2/commands/security/verbs/fixtures/client_service_node.py
```diff
@@ -0,0 +1,55 @@
...
+
+    node = ClientServiceNode()
+
+    try:
+        rclpy.spin(node)
+    except KeyboardInterrupt:
+        print('node stopped cleanly')
+    except BaseException:
+        print('exception in node:', file=sys.stderr)
+        raise
```
hidmic: @BMarchi nit: isn't this `except` clause somewhat redundant?

sros2/test/sros2/commands/security/verbs/test_generate_policy.py
```diff
@@ -12,116 +12,185 @@
...
+            assert len(
+                [s for s in services if s.text == 'test_generate_policy_services_client']) == 1
+            assert len(
+                [s for s in services if s.text == 'test_generate_policy_services_server']) == 0
+
+    @launch_testing.markers.retry_on_failure(times=2)
+    def test_generate_policy_no_policy_file(self):
+        with self.launch_gen_policy_command(arguments=[]) as gen_command:
+            assert gen_command.wait_for_shutdown(timeout=10)
+        assert gen_command.exit_code != 0
```
hidmic: @BMarchi nit:
```suggestion
        assert gen_command.exit_code != launch_testing.asserts.EXIT_OK
```

hidmic:
Blocked on https://github.com/ros2/ros2cli/issues/419.

hidmic:
Rebased, refactored and used as a basis for #214. Closing.

PR #174, opened 2020-01-18
switch to not deprecated API

dirk-thomas:
Related to ros2/ros2cli#436.

### DIFF
--- a/sros2/sros2/command/security.py
+++ b/sros2/sros2/command/security.py
@@ -12,9 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Technically this one will need a minimum version dependency on the next version of `ros2cli`.

mikaelarguedas:
This appears to have broken all the end-to-end tests https://ci.ros2.org/view/nightly/job/nightly_linux_release/1422/testReport/junit/sros2.test.sros2.commands.security.verbs/

mjcarroll:
It has, and @dirk-thomas is looking into it.

PR #177, opened 2020-03-15
Use Contexts

ruffsl:
Related: https://github.com/ros2/design/pull/274#issuecomment-598892231

#### TODO:

- [x] Update policy schema with contexts
- [x] Update policy transforms for DDS
- [x] Update keystore layout with pubic, private, contexts sub folders
- [x] Update existing tests and downstream packages

#### Bonus items
- [ ] Update auto generation
  - would require context aware telemetry from rosgraph API
  - disable otherwise if rosgraph API doesn't expose context info
- [ ] Update documentation
- [ ] Support for metadata and `<profiles>` type attribute
  - e.g. associating profiles to domains, validity times, singing CA
  - may need more thought, could look back at keymint for config patterns


ruffsl:
Ok, I think the basics are working at least:

<details>

```
$ ros2 security generate_artifacts -k keystore -p ~/ws/ros2/src/ros2/sros2/sros2/test/policies/talker_listener.policy.xml 
keystore is not a valid keystore, creating new keystore
creating keystore: keystore
creating new CA key/cert pair
creating governance file: keystore/contexts/governance.xml
creating signed governance file: keystore/contexts/governance.p7s
all done! enjoy your keystore in keystore
cheers!
creating key for identity: '/talker_listener/talker'
creating cert and key
creating permission file for identity: '/talker_listener/talker'
creating key for identity: '/talker_listener/listener'
creating cert and key
creating permission file for identity: '/talker_listener/listener'

$ tree keystore/
keystore/
├── contexts
│   ├── governance.p7s
│   ├── governance.xml
│   └── talker_listener
│       ├── listener
│       │   ├── cert.pem
│       │   ├── governance.p7s -> ../../governance.p7s
│       │   ├── identity_ca.cert.pem -> ../../../public/identity_ca.cert.pem
│       │   ├── key.pem
│       │   ├── permissions.p7s
│       │   ├── permissions.xml
│       │   └── permissions_ca.cert.pem -> ../../../public/permissions_ca.cert.pem
│       └── talker
│           ├── cert.pem
│           ├── governance.p7s -> ../../governance.p7s
│           ├── identity_ca.cert.pem -> ../../../public/identity_ca.cert.pem
│           ├── key.pem
│           ├── permissions.p7s
│           ├── permissions.xml
│           └── permissions_ca.cert.pem -> ../../../public/permissions_ca.cert.pem
├── private
│   ├── ca.key.pem
│   ├── identity_ca.key.pem -> ca.key.pem
│   └── permissions_ca.key.pem -> ca.key.pem
└── public
    ├── ca.cert.pem
    ├── identity_ca.cert.pem -> ca.cert.pem
    └── permissions_ca.cert.pem -> ca.cert.pem

6 directories, 22 files
```

</details>

ping @mikaelarguedas @kyrofa @ivanpauno

sros2/sros2/policy/defaults/policy.xml
```diff
@@ -1,16 +1,21 @@
...
-      </topics>
-      <services reply="ALLOW" request="ALLOW">
-        <service>/*</service>
-      </services>
-      <actions call="ALLOW" execute="ALLOW">
-        <action>/*</action>
-      </actions>
-    </profile>
-  </profiles>
+<policy version="0.1.0"
```
ivanpauno: this version here was the version of the policy file format?
should it be updated?

sros2/test/policies/add_two_ints.policy.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
ivanpauno: minimal note: It would be good to not squash the commit renaming files before merging, to have a cleaner history.

sros2/sros2/policy/defaults/policy.xml
mikaelarguedas: policy version will need to be bumped

sros2/test/policies/permissions/single_context/permissions.xml
```diff
@@ -0,0 +1,195 @@
+<dds xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://www.omg.org/spec/DDS-SECURITY/20170901/omg_shared_ca_permissions.xsd">
```
mikaelarguedas: Why are these files committed? Will they be used in the tests?
Or is it just to show what the resulting permission files would be ?

sros2/sros2/policy/schemas/policy.xsd
```diff
@@ -10,15 +10,32 @@
...
+    </xs:complexType>
+
     <xs:complexType name="Profiles">
         <xs:sequence minOccurs="1" maxOccurs="unbounded">
             <xs:element name="profile" type="Profile" />
         </xs:sequence>
+        <!-- <xs:sequence minOccurs="0" maxOccurs="1">
+            <xs:element name="metadata" type="xsd:anyType" />
+        </xs:sequence>
+        <xs:attribute name="type" type="xs:string" use="required" /> -->
```
mikaelarguedas: Even if we don't get around *using* the metadata in the tool by foxy, it would be great to get the new tag in the policy schema (so that any change in the tool doesn't require us to bump the version of the schema just for the metadata tag)

ivanpauno:
@ruffsl To allow publishing/subscribing to `ros_discovery_info` topic, I pushed a [branch](https://github.com/ros2/sros2/tree/ivanpauno/contexts) with two commits:
- The [first commit](https://github.com/ros2/sros2/commit/8bf8974aa0e5a41eab8e1274b41e7d92298df40a) adds a global parameter in the permissions dds template, to add or not that ALLOW rule.
- The [second commit](https://github.com/ros2/sros2/commit/87dcaf559f397b91e51b57d4ffadad9ae4241cde), sets the parameter to `true` if the detected rmw implementation is `fastrtps` based. It uses `rclpy` `get_rmw_implementation_identifier` function if available, if not checks the `RMW_IMPLEMENTATION` environment variable (which can be set in environments without ROS2 installed).

I didn't open a PR to avoid adding extra noise and just in case you want to implement this differently. The first commit can probably be taken as-is. The second one, you maybe want to add something more explicit to the verbs arguments.

>  Update existing tests and downstream packages

I have already updated [system_tests/test_security](https://github.com/ros2/system_tests/pull/406) based on changes here and in `rcl`.

> would require context aware telemetry from rosgraph API

I have already added this to fastrtps, and connext rmw implementations. I have to add something similar in cyclonedds (though they don't support security yet, just for completion), and implement the function in `rcl`, `rclpy`. It would be great to complete this in a follow-up though, to limit a bit the size of the changes.

ivanpauno:
@ruffsl Friendly ping.
It would be great to wrap up this for the beginning of next week (e.g. Wednesday), considering that Foxy deadline is close.
I can help with some of the remaining tasks if you don't have time to handle them (e.g.: updating the tests). Let me know if I can help with anything.

sros2/test/policies/permissions/single_context/permissions.xml
ruffsl: I checked them in, as it was helpful to track what alterations in the final permission file would occur when modifying the template used for various policy formations: e.g.  c4b20e3f1bf8ebaaa9902be078ae40b6e859efcc

It would also be nice to also just reuse there for unit testing for further transforms checks.

sros2/sros2/policy/schemas/policy.xsd
ruffsl: I added this back, but wasn't sure of how we want to full integrate the `type` attribute, so I switched it to `use="optional"`.

sros2/sros2/api/__init__.py
```diff
@@ -226,12 +247,16 @@ def is_key_name_valid(name):
...
     permissions_xsl_path = get_transport_template('dds', 'permissions.xsl')
     permissions_xsl = etree.XSLT(etree.parse(permissions_xsl_path))
     permissions_xsd_path = get_transport_schema('dds', 'permissions.xsd')
     permissions_xsd = etree.XMLSchema(etree.parse(permissions_xsd_path))
 
-    permissions_xml = permissions_xsl(policy_element)
+    kwargs = {}
+    if get_current_rmw_implementation() in ('rmw_fastrtps_cpp','rmw_fastrtps_dynamic_cpp'):
+        kwargs['allow_ros_discovery_topic'] = etree.XSLT.strparam("1")
+    permissions_xml = permissions_xsl(policy_element, **kwargs)
```
ruffsl: Whoa, that cool! Didn't know xsl templates could take in `kwargs`, TDIL.

sros2/sros2/policy/templates/dds/permissions.xsl
```diff
@@ -57,6 +59,21 @@
...
+                      <topic>ros_discovery_info</topic>
+                    </topics>
+                  </publish>
+                  <subscribe>
+                    <topics>
+                      <topic>ros_discovery_info</topic>
+                    </topics>
+                  </subscribe>
+                </allow_rule>
+              </xsl:if>
```
ruffsl: Lets hope we don't have to keep this fully connected data flow rule forever.

ruffsl:
Hey @ivanpauno @mikaelarguedas , I had a moment to push some commits to address things.

> The first commit can probably be taken as-is. The second one, you maybe want to add something more explicit to the verbs arguments.

I've cherry picked them into the PR. You're suggesting adding a verb to control this? If its already reading from the env, we may not need to clutter the CLI with a subsumption layer of settings.

> It would be great to complete this in a follow-up though, to limit a bit the size of the changes.

Sounds fine.

> I can help with some of the remaining tasks if you don't have time to handle them (e.g.: updating the tests). Let me know if I can help with anything.

Thanks! Could you wrap up this PR with the tests and any hanging review comments. I can't immediately think of a pythonic pattern to do the file/dir/symlink checks for the keystore, and I've got another deadline for my lab this week.

sros2/sros2/verb/create_key.py
```diff
@@ -26,10 +26,15 @@ class CreateKeyVerb(VerbExtension):
     """Create key."""
 
     def add_arguments(self, parser, cli_name):
-        arg = parser.add_argument('ROOT', help='root path of keystore')
+        arg = parser.add_argument(
+            '-k', '--keystore-root-path',
```
ivanpauno: To avoid a hard API break, 'ROOT' can be maintained too.
It could be also detected if it was used and print a deprecation warning if later is going to be removed.

P.S.: People will have to read about changes in `sros2` for Foxy, so maybe it's not too bad to break this too.

sros2/sros2/verb/create_key.py
```diff
@@ -26,10 +26,15 @@ class CreateKeyVerb(VerbExtension):
...
+            help='root path of keystore')
         arg.completer = DirectoriesCompleter()
-        parser.add_argument('NAME', help='key name, aka ROS node name')
+        parser.add_argument(
+            '-c', '--context', nargs='*', default=[],
```
ivanpauno: similar comment, `NAME` could be kept.

sros2/sros2/verb/generate_artifacts.py
```diff
@@ -31,11 +31,13 @@ class GenerateArtifactsVerb(VerbExtension):
...
     def add_arguments(self, parser, cli_name):
-        arg = parser.add_argument('-k', '--keystore-root-path', help='root path of keystore')
+        arg = parser.add_argument(
+            '-k', '--keystore-root-path',
+            help='root path of keystore')
         arg.completer = DirectoriesCompleter()
         parser.add_argument(
-            '-n', '--node-names', nargs='*', default=[],
-            help='list of identities, aka ROS node names')
+            '-c', '--contexts', nargs='*', default=[],
```
ivanpauno: `-n` could also be kept. It should be interpreted as "names" (i.e.: "security contexts names").

sros2/sros2/verb/generate_artifacts.py
```diff
@@ -31,11 +31,13 @@ class GenerateArtifactsVerb(VerbExtension):
...
+            help='list of identities, aka ROS context paths')
```
ivanpauno: we could be more explicit, and use `security context` instead of `context`.

sros2/sros2/api/__init__.py
ivanpauno: Me neither :rofl:. I liked it, because it avoids duplicating the template (with minimal changes).

sros2/sros2/verb/generate_artifacts.py
ruffsl: Ok, `--security-contexts` could work, matching the rcl args.

sros2/sros2/policy/templates/dds/permissions.xsl
ivanpauno: Yes. I think that in a near future (one/two months), all the `rmw` implementation will be migrated.

ivanpauno:
> Thanks! Could you wrap up this PR with the tests and any hanging review comments. I can't immediately think of a pythonic pattern to do the file/dir/symlink checks for the keystore, and I've got another deadline for my lab this week.

Yes, I will follow up with this tomorrow. Thanks for your prompt answer!!

sros2/sros2/verb/generate_artifacts.py
ruffsl: Context paths do not have an equivalent meaning as node names, and I think the args should conflate the two to make that distinction clear, so the use recognizes foxy has shifted to context.

sros2/sros2/verb/create_key.py
ruffsl: > People will have to read about changes in sros2 for Foxy, so maybe it's not too bad to break this too.

Yeah, I wanted to use this as an opportunity to make the CLI more consistent. The mix use of positional and keyed command arguments was really annoying in terms if UI.

sros2/sros2/verb/generate_artifacts.py
ivanpauno: Sounds reasonable, applies to https://github.com/ros2/sros2/pull/177#discussion_r394644744 too.

sros2/sros2/verb/generate_artifacts.py
ivanpauno: I was referring to the help comment, but updating the long flag version sounds good too.

sros2/sros2/verb/create_key.py
kyrofa: Yeah we're talking about cleaning up the non-CLI API for Foxy anyway.

sros2/sros2/policy/templates/dds/permissions.xsl
mikaelarguedas: I believe what @ruffsl mean is that hopefully this will be done out of band and not require a fully connected rule on this topic.

sros2/sros2/verb/create_key.py
mikaelarguedas: :+1: I was considering doing this before Foxy as well, currently only newly introduced verbs like `generate_artifacts` have keyed arguments. For clarity it would be better to have them in a separate PR. If we keep it in this one, can we make this in an independent commit (similar to file renaming commits) ?

sros2/sros2/api/__init__.py
```diff
@@ -294,38 +334,39 @@ def create_key(keystore_path, identity):
...
                 raise RuntimeError(str(e))
 
-    # copy the governance file in there
-    keystore_governance_path = os.path.join(keystore_path, 'governance.p7s')
+    # symlink the governance file in there
+    keystore_governance_path = os.path.join(keystore_path, KS_CONTEXT, 'governance.p7s')
     dest_governance_path = os.path.join(key_dir, 'governance.p7s')
-    shutil.copyfile(keystore_governance_path, dest_governance_path)
+    relativepath = os.path.relpath(keystore_governance_path, key_dir)
+    os.symlink(src=relativepath, dst=dest_governance_path)
```
mikaelarguedas: What happens in the destination file already exists?

sros2/sros2/policy/templates/dds/permissions.xsl
```diff
@@ -20,52 +20,75 @@
...
-                  <xsl:call-template name="TranslatePermissions">
-                    <xsl:with-param name="qualifier" select="'ALLOW'"/>
-                  </xsl:call-template>
-                </xsl:for-each>
-              </allow_rule>
-            </xsl:if>
-            <default>DENY</default>
-          </grant>
+          <xsl:for-each select="profiles">
+            <grant name="{$common_name}">
```
mikaelarguedas: Does it actually work to have different grant names for nodes in the same context ?
it seems to me that a single grant would be used for all nodes in the participant.

there seem to be now 3 possible permission files generated:
- permissions files with single grant with all the permissions of the participant (`/single_context`)
- permissions files with single grant for unique node ran standalone
- permissions files with multiple grants, each encompassing a single node (e.g. `permissions/talker_listener/permissions.xml`)
How do you chose to generate the second or the third one ?
In what use case should the one with multiple grants be used?

sros2/sros2/policy/templates/dds/permissions.xsl
ivanpauno: Oh, I see. I think it's definitely possible.
It would be great to finish discussing this alternative in the design document, before the other migrations are done.
Of course, we can first migrate all to use the `ros_discovery_info` topic, and then change them again to use `Participant` `userData`.

sros2/sros2/verb/create_key.py
ivanpauno: I will consolidate things in different commits before merging.

sros2/sros2/verb/generate_artifacts.py
ivanpauno: Solved in https://github.com/ros2/sros2/pull/177/commits/d6b2936845f69c66d1b6347c7783f82bc5379384

sros2/sros2/api/__init__.py
```diff
@@ -198,17 +242,21 @@ def is_key_name_valid(name):
     try:
         return (validate_namespace(node_ns) and validate_node_name(node_name))
```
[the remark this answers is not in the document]
ivanpauno: I've created a function in `rcl`: `rcl_validate_security_context_name`.
It does pretty much the same that `rmw_validate_full_topic_name`.

I will use the later, until I propagate the `rcl` function to `rclpy`.

sros2/sros2/api/__init__.py
ivanpauno: If the file already exists, it will raise a `FileExistsError`.
Should the old file be deleted and overridden?

sros2/sros2/policy/schemas/policy.xsd
```diff
@@ -10,15 +10,34 @@
     <xs:element name="policy" type="Policy" />
     <xs:complexType name="Policy">
         <xs:sequence minOccurs="1" maxOccurs="1">
-            <xs:element name="profiles" type="Profiles" />
+            <xs:element name="contexts" type="Contexts" />
         </xs:sequence>
         <xs:attribute name="version" type="xs:string" use="required" />
```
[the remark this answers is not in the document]
ivanpauno: SGTM, done in https://github.com/ros2/sros2/pull/177/commits/11159158f282cae846da45185c5cf8ffe2854bdb.

sros2/sros2/api/__init__.py
```diff
@@ -143,22 +143,35 @@ def create_governance_file(path, domain_id):
...
+        os.makedirs(os.path.join(keystore_path, 'private'), exist_ok=True)
+        os.makedirs(os.path.join(keystore_path, 'contexts'), exist_ok=True)
     else:
-        print('directory already exists: %s' % keystore_path)
-
-    ca_key_path = os.path.join(keystore_path, 'ca.key.pem')
-    ca_cert_path = os.path.join(keystore_path, 'ca.cert.pem')
-
-    if not (os.path.isfile(ca_key_path) and os.path.isfile(ca_cert_path)):
+        print('keystore already exists: %s' % keystore_path)
```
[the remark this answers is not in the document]
ivanpauno: Fixed in https://github.com/ros2/sros2/pull/177/commits/0c9d291362ccbcaed61a238592f179b95825f847.

sros2/sros2/api/__init__.py
```diff
@@ -143,22 +143,35 @@ def create_governance_file(path, domain_id):
...
+
+    keystore_ca_cert_path = os.path.join(keystore_path, 'public', 'ca.cert.pem')
+    keystore_ca_key_path = os.path.join(keystore_path, 'private', 'ca.key.pem')
+    
+    keystore_permissions_ca_cert_path = os.path.join(keystore_path, 'public', 'permissions_ca.cert.pem')
+    keystore_permissions_ca_key_path = os.path.join(keystore_path, 'private', 'permissions_ca.key.pem')
+    keystore_identity_ca_cert_path = os.path.join(keystore_path, 'public', 'identity_ca.cert.pem')
+    keystore_identity_ca_key_path = os.path.join(keystore_path, 'private', 'identity_ca.key.pem')
+
+    if not (os.path.isfile(keystore_permissions_ca_cert_path) and os.path.isfile(keystore_permissions_ca_key_path) and
```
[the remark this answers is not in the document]
ivanpauno: Done in https://github.com/ros2/sros2/pull/177/commits/1c952524f44875d2a40e9f544607cf8bf2f937f2.

ivanpauno:
I've addressed comments and made tests pass.
Though I will do some more updates tomorrow, this is ready to be reviewed again.

sros2/sros2/api/__init__.py
```diff
@@ -364,16 +381,26 @@ def create_key(keystore_path, identity):
...
 
 def list_keys(keystore_path):
-    for name in os.listdir(keystore_path):
-        if os.path.isdir(os.path.join(keystore_path, name)):
+    contexts_path = os.path.join(keystore_path, KS_CONTEXT)
+    if not os.path.exists(keystore_path):
+        raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), keystore_path)
+    if not os.path.exists(contexts_path):
+        return True
+    for name in os.listdir(contexts_path):
```
ruffsl: Context paths can be nested. Is listdir recursive?
I.e. will it walk over contexts that aren't just in flat in the root of the context sub folder?

sros2/sros2/policy/templates/dds/permissions.xsl
ruffsl: > Does it actually work to have different grant names for nodes in the same context ?

Correct me if I coded this wrong, but here a `<profiles>` tag (plural) must map to a grant, as it is the containing child element that would allow the parent `<context>` tag to have a sequence of unique non-overlapping domain id rules. The `<profile>` tag (singular) is the only element that one-to-one maps to a individual node.

The transform should do all the dds mangling on mass in one dds permission xml tree, then the sros2 api should just find/extract the `grant`(s) with the `name` attribute that match the context path or identity requested.

Could you give/link example inputs/outputs for the three possibilities, so I understand?

mikaelarguedas: The code seems to do what I expect it to (aka 1 or 2 based on the policy file provided).
The thing I don't get is 3:
an exampel of such files is this one committed in the PR https://github.com/ros2/sros2/blob/1a6048818002c45e5decd69ded436b41e6e20e73/sros2/test/policies/permissions/talker_listener/permissions.xml
How can they be generated by the CLI?
When are they expected to be used in practice?

My current understanding is that this permission file is generated by `policy_to_permissions.py` for the sole purpose of testing but cannot be generated via the CLI. Is that correct ?

sros2/sros2/verb/create_key.py
```diff
@@ -26,10 +26,16 @@ class CreateKeyVerb(VerbExtension):
...
 
     def add_arguments(self, parser, cli_name):
-        arg = parser.add_argument('ROOT', help='root path of keystore')
+        arg = parser.add_argument(
+            '-k', '--keystore-root-path',
+            help='root path of keystore')
         arg.completer = DirectoriesCompleter()
-        parser.add_argument('NAME', help='key name, aka ROS node name')
+        parser.add_argument(
+            '-c', '--security-contexts', nargs='*', default=[],
```
mikaelarguedas: this changes the semantic of the verb.
Before this was creating a single key, now it is taking a list of security context paths and creating them all.

Either behavior should be kept or the name updated to match the new behavior (the latter sounds good to me)

sros2/sros2/verb/create_permission.py
```diff
@@ -31,17 +31,23 @@ class CreatePermissionVerb(VerbExtension):
...
 
     def add_arguments(self, parser, cli_name):
-        arg = parser.add_argument('ROOT', help='root path of keystore')
+        arg = parser.add_argument(
+            '-k', '--keystore-root-path',
+            help='root path of keystore')
         arg.completer = DirectoriesCompleter()
-        parser.add_argument('NAME', help='key name, aka ROS node name')
+        parser.add_argument(
+            '-c', '--security-contexts', nargs='*', default=[],
```
mikaelarguedas: Same as above this is changing behavior and semantic

sros2/sros2/verb/create_permission.py
```diff
@@ -31,17 +31,23 @@ class CreatePermissionVerb(VerbExtension):
...
+            help='identity, aka ROS security contexts path')
         arg = parser.add_argument(
-            'POLICY_FILE_PATH', help='path of the policy xml file')
+            '-p', '--policy-files', nargs='*', default=[],
```
mikaelarguedas: This seems wrong. AFAICT `create_permission` takes a single policy file

sros2/sros2/verb/generate_artifacts.py
```diff
@@ -45,7 +47,7 @@ def add_arguments(self, parser, cli_name):
     def main(self, *, args):
         try:
             success = generate_artifacts(
-                args.keystore_root_path, args.node_names, args.policy_files)
+                args.keystore_root_path, args.contexts, args.policy_files)
```
mikaelarguedas:
```suggestion
                args.keystore_root_path, args.security_contexts, args.policy_files)
```

sros2/sros2/api/__init__.py
ivanpauno: > To be fair it was already the case before this change so doesn't have to be addressed here. Should be ticketed though if not addressed

Agreed.
What should be the output of:

```
-> contexts
  -> context1
    -> security artifacts ...
    -> nested1
      -> security artifacts ...
  ->context2
    ->nested2
      ->security artifacts...
```
?

IMO, it should be:

```
/context1
/context1/nested1
/context2/nested2
```

If the `contexts` folder have valid security artifacts, `/` should be showed too.

sros2/sros2/api/__init__.py
```diff
@@ -40,6 +40,11 @@
     load_policy,
 )
 
+try:
+    from rclpy.utilities import get_rmw_implementation_identifier
+except ImportError:
```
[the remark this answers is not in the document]
ivanpauno: If `rclpy` is not available.
It may not make much sense now, as `rclpy` is a required dependency.

I can delete that check for now, and it can be re-added in the future when getting rid of `rclpy` dependency.

sros2/sros2/api/__init__.py
```diff
@@ -294,45 +346,61 @@ def create_key(keystore_path, identity):
...
+    )
 
     return True
 
 
 def list_keys(keystore_path):
-    for name in os.listdir(keystore_path):
-        if os.path.isdir(os.path.join(keystore_path, name)):
+    contexts_path = os.path.join(keystore_path, KS_CONTEXT)
+    if not os.path.exists(keystore_path):
```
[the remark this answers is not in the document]
ivanpauno: `isdir` is a better check, thanks!

sros2/sros2/api/__init__.py
mikaelarguedas: Sounds good to me

sros2/sros2/api/__init__.py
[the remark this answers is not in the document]
kyrofa: Seems sensible to let the error propagate if rclpy is a required dep.

sros2/sros2/verb/generate_artifacts.py
ivanpauno: Addressed in https://github.com/ros2/sros2/pull/177/commits/103e18c1366ed793c6269f2d9adab8a16aaf7046.

sros2/sros2/verb/create_permission.py
ivanpauno: Addressed in https://github.com/ros2/sros2/pull/177/commits/103e18c1366ed793c6269f2d9adab8a16aaf7046.

sros2/sros2/verb/create_permission.py
ivanpauno: Addressed in https://github.com/ros2/sros2/pull/177/commits/103e18c1366ed793c6269f2d9adab8a16aaf7046.

sros2/sros2/verb/create_key.py
ivanpauno: Addressed in https://github.com/ros2/sros2/pull/177/commits/103e18c1366ed793c6269f2d9adab8a16aaf7046.

sros2/sros2/api/__init__.py
```diff
@@ -142,23 +167,46 @@ def create_governance_file(path, domain_id):
...
+        keystore_identity_ca_cert_path,
+        keystore_identity_ca_key_path,
+    )
 
-    if not (os.path.isfile(ca_key_path) and os.path.isfile(ca_cert_path)):
+    if not all(os.path.isfile(x) for x in required_files):
         print('creating new CA key/cert pair')
-        create_ca_key_cert(ca_key_path, ca_cert_path)
+        create_ca_key_cert(keystore_ca_key_path, keystore_ca_cert_path)
+        os.symlink(src='ca.cert.pem', dst=keystore_permissions_ca_cert_path)
```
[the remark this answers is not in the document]
ivanpauno: Addressed in https://github.com/ros2/sros2/pull/177/commits/f5080e8bfa4ea0ec88d5db87d7dd23ba841b9fb1.

sros2/sros2/api/__init__.py
ivanpauno: Addressed in https://github.com/ros2/sros2/pull/177/commits/f5080e8bfa4ea0ec88d5db87d7dd23ba841b9fb1.

Instead of `delete + override`, I'm raising an error in that last case.
I'm afraid to delete something a distracted user didn't want.

sros2/sros2/api/__init__.py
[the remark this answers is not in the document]
ivanpauno: Done in https://github.com/ros2/sros2/pull/177/commits/51405d63f195e0f9917a687b0d2b13825c679018.

sros2/sros2/policy/templates/dds/permissions.xsl
ruffsl: In the file you linked, the `<grant>` tags have different `name` attributes, a thus derive from separate `context` elements in a `policy`. The file is what the raw xml tree looks like when returned by the xsl transform.

> but cannot be generated via the CLI. Is that correct ?

Yes, there isn't a way that users could accidentally misuse the CLI to generate a permission for a single context that would include grants from other contexts, as that would just defeat the purpose of separation of privilege/contexts.

sros2/sros2/verb/create_permission.py
```diff
@@ -31,17 +31,17 @@ class CreatePermissionVerb(VerbExtension):
...
+        arg = parser.add_argument('-k', '--keystore-root-path', help='root path of keystore')
         arg.completer = DirectoriesCompleter()
-        parser.add_argument('NAME', help='key name, aka ROS node name')
-        arg = parser.add_argument(
-            'POLICY_FILE_PATH', help='path of the policy xml file')
-        arg.completer = FilesCompleter(
-            allowednames=('xml'), directories=False)
+        parser.add_argument(
+            '-c', '--security-context', help='identity, aka ROS security contexts path')
+        arg = parser.add_argument('-p', '--policy-files', help='list of policy xml file paths')
```
mikaelarguedas: This is still misleading IMO as the underlying function takes a single policy file.. Can you please use the original help text and a singular form for the long flag name ?

sros2/sros2/policy/templates/dds/permissions.xsl
mikaelarguedas: > Yes, there isn't a way that users could accidentally misuse the CLI to generate a permission.xml file for a single context that would include grants from other contexts, as that would just defeat the purpose of separation of privilege/contexts.

:+1: this is why I was confused to see such files committed.
As these cannot be obtained by users from the CLI that sounds good :+1: thanks for clarifying.

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2020-03-18/13313/1

sros2/sros2/verb/create_permission.py
ivanpauno: Sorry, see https://github.com/ros2/sros2/pull/177/commits/c23bd7b43db0b51bbe387b71124200a60a570fda.

sros2/sros2/api/__init__.py
mikaelarguedas: Maybe it's my workflow but I find it convenient to change a couple rules in a policy file and just rerun `generate_artifacts` on my entire policy file and trust the tool to override the changed permission files accordingly. It looks like with this more conservative approach I would need to go prune by hand all the directories that may have a permission change.

I'd prefer not changing the previous behavior, but either way it would be good to provide an option to opt-in (or out) of overwriting existing files

ruffsl: In my prior work with keymint, I treated auto generated security files as build artifacts, and thus subsequent build file could be modified at will by the build tool, where the user would recognize the temporal nature of these files if they instructed the build tool to rebuild changed source files. So I'd say let the tool overwrite it the generated files by default.

In keymint, I also separated the pipeline into a src. build, and install folder structure, so we may want to explore using similar workflows that could separate the policy source files, from the generated intermediate representations built, to the final signed security artifacts installed. Keymint had a `--dry-run` flag that the CLI could check to see if it should warn about detected file diffs and not write to disk.

ivanpauno: Done in https://github.com/ros2/sros2/pull/177/commits/d52333215652ccce161473dbf722a3d4a40c5300.

ivanpauno:
@ruffsl @mikaelarguedas 
I'm planning to do the following items in a follow-up:
- Update auto generation
- Update documentation
- https://github.com/ros2/sros2/pull/177#discussion_r395352292

Something else missing to get this approved?

ruffsl:
I was only wondering about that last unresolved comment you linked. If you like to follow up that with a septate PR, that fine. This repo is set so authors of PR can't approve the same PR, but this all LGTM!

sros2/setup.py
```diff
@@ -66,7 +66,9 @@ def package_files(directory):
             ':CreatePermissionVerb',
             'distribute_key = sros2.verb.distribute_key:DistributeKeyVerb',
             'generate_artifacts = sros2.verb.generate_artifacts:GenerateArtifactsVerb',
-            'generate_policy = sros2.verb.generate_policy:GeneratePolicyVerb',
+            # TODO(ivanpauno): Reactivate this after having a way to instronspect
+            # security context names in rclpy
```
kyrofa: Do we have an issue in rclpy that we can use for tracking this?

sros2/sros2/api/__init__.py
```diff
@@ -181,34 +221,35 @@ def create_keystore(keystore_path):
...
 def create_permission_file(path, domain_id, policy_element):
+    print('creating permission')
     permissions_xsl_path = get_transport_template('dds', 'permissions.xsl')
     permissions_xsl = etree.XSLT(etree.parse(permissions_xsl_path))
     permissions_xsd_path = get_transport_schema('dds', 'permissions.xsd')
     permissions_xsd = etree.XMLSchema(etree.parse(permissions_xsd_path))
 
-    permissions_xml = permissions_xsl(policy_element)
+    kwargs = {}
+    if get_rmw_implementation_identifier() in ('rmw_fastrtps_cpp', 'rmw_fastrtps_dynamic_cpp'):
```
kyrofa: I assume this list will grow over time, yes? I would be easier to maintain if we could extract this into a non-public global here instead of maintaining the list here in the middle of the code.

sros2/test/sros2/commands/security/verbs/test_create_keystore.py
```diff
@@ -31,24 +31,41 @@ def keystore_dir(tmpdir_factory):
...
-    assert len(os.listdir(keystore_dir)) == len(expected_files)
 
-    for expected_file in expected_files:
-        assert os.path.isfile(os.path.join(keystore_dir, expected_file))
+    assert len(list(keystore_dir.iterdir())) == 3
+    assert len(list(public.iterdir())) == len(expected_files_public)
+    assert len(list(private.iterdir())) == len(expected_files_private)
+    assert len(list(contexts.iterdir())) == len(expected_files_contexts)
+    expected_files = expected_files_public + expected_files_private + expected_files_contexts
+    assert all(x.exists() for x in expected_files)
```
mikaelarguedas: Nit: use `isfile` instead of `exists`

sros2/test/sros2/commands/security/verbs/test_generate_policy.py
```diff
@@ -24,6 +24,8 @@
 from std_srvs.srv import Trigger
 
 
+# TODO(ivanpauno): reactivate this test after updating generate policy
+@pytest.mark.skipif(True, reason='temporaly deacticated')
```
mikaelarguedas:
```suggestion
@pytest.mark.skipif(True, reason='temporarily deactivated')
```
Same multiple times below

sros2_cmake/README.md
```diff
@@ -1,19 +1,21 @@
...
+    # ros2_security_generate_artifacts(SECURITY_CONTEXTS <context_1> <context_2>...<context_n>)
 
-    # NODES (macro multi-arg) takes the node names for which keys will be generated
+    # SECURITY_CONTEXTS (macro multi-arg) takes the security contexts names for which keys will be generated
+    #   Each executable can be using a different (or the same), security contexts.
+    #   All nodes in the same process use the same security context.
     # SECURITY (cmake arg) if not define or OFF, will not generate key/keystores
     # ROS_SECURITY_ROOT_DIRECTORY (env variable) the location of the keystore
-    # POLICY_FILE (cmake arg) if defined, will compile policies by node name into the access private certificates (e.g POLICY_FILE=/etc/policies/<policy.xml>, Generate: <node_name> /etc/policies/<policy.xml>) **if defined, all nodes must have a policy defined for them**
+    # POLICY_FILE (cmake arg) if defined, will compile policies by security context name into the access private certificates (e.g POLICY_FILE=/etc/policies/<policy.xml>, Generate: <context_name> /etc/policies/<policy.xml>) **if defined, all contexts must have a policy defined for them**
```
mikaelarguedas: Unrelated to this PR but I can't comprehend this sentence

sros2_cmake/cmake/ros2_security_generate_artifacts.cmake
```diff
@@ -12,12 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-macro(ros2_secure_node)
-  # ros2_secure_node(NODES <node_1> <node_2>...<node_n>)
+macro(ros2_security_generate_artifacts)
```
mikaelarguedas: Nit: this is a mouthful

How about `sros2_generate_artifacts`? this would convey information about where the macro comes from as well as which sros2 verb it's linked to

sros2/setup.py
mikaelarguedas: +1, TODOs in code usually don't get revisited if there isn't an associated ticket (same for other TODOs below)

Spelling: instronspect -> introspect

ivanpauno: See https://github.com/ros2/sros2/issues/182, https://github.com/ros2/rclpy/issues/528, https://github.com/ros2/rclpy/issues/529, https://github.com/ros2/sros2/issues/183.

sros2_cmake/README.md
ivanpauno: Yeah, it's hard to read.
Proposed rephrasing:
```
if defined, will generate permissions based on the policies defined for each context in the file. All contexts must have a policy defined for them
```

sros2_cmake/cmake/ros2_security_generate_artifacts.cmake
ivanpauno: Sounds good!

ivanpauno: Done in https://github.com/ros2/sros2/pull/177/commits/2db3d4e8bff1b56e6621b170a78feeafcd4b1fbd.

sros2/test/sros2/commands/security/verbs/test_generate_policy.py
ivanpauno: Ups, thanks!
Done between https://github.com/ros2/sros2/pull/177/commits/2db3d4e8bff1b56e6621b170a78feeafcd4b1fbd and https://github.com/ros2/sros2/pull/177/commits/339b1390e9f105e69f92181c4123f5617375dbbe.

sros2/test/sros2/commands/security/verbs/test_create_keystore.py
ivanpauno: Done, see https://github.com/ros2/sros2/pull/177/commits/2db3d4e8bff1b56e6621b170a78feeafcd4b1fbd.

sros2/test/sros2/commands/security/verbs/test_generate_policy.py
```diff
@@ -24,6 +24,8 @@
 from std_srvs.srv import Trigger
 
 
+# TODO(ivanpauno): reactivate this test after updating generate policy
+@pytest.mark.skipif(True, reason='temporarily deactivated')
```
hidmic: @ivanpauno FYI `pytest.mark.skip` has the same effect, unconditionally.

kyrofa:
Heads up, this has a small conflict.

sros2_cmake/README.md
mikaelarguedas: Yeah too many keywords, there is actually a single policy, that contains contexts, that contain profiles. A simpler version to consider:

```
if defined, will generate security artifacts for each context defined in the policy file.
```

The last bit about "must have" is actually not true since https://github.com/ros2/sros2/pull/108

sros2/sros2/api/__init__.py
ivanpauno: Done in https://github.com/ros2/sros2/pull/177/commits/2db3d4e8bff1b56e6621b170a78feeafcd4b1fbd.

sros2/test/sros2/commands/security/verbs/test_generate_policy.py
ivanpauno: Done in https://github.com/ros2/sros2/pull/177/commits/070105b15778030a10ff45a98d02964689752822.

sros2_cmake/README.md
```diff
@@ -1,19 +1,21 @@
...
+`sros2_generate_artifacts(SECURITY_CONTEXTS <context_name>)`  
 
 Macro definition:  
 ```
-    # ros2_secure_node(NODES <node_1> <node_2>...<node_n>)
+    # sros2_generate_artifacts(SECURITY_CONTEXTS <context_1> <context_2>...<context_n>)
 
-    # NODES (macro multi-arg) takes the node names for which keys will be generated
+    # SECURITY_CONTEXTS (macro multi-arg) takes the security contexts names for which keys will be generated
+    #   Each executable can be using a different (or the same), security contexts.
```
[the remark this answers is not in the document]
ivanpauno: See https://github.com/ros2/sros2/pull/177/commits/070105b15778030a10ff45a98d02964689752822.

sros2/sros2/verb/create_key.py
```diff
@@ -26,10 +26,13 @@ class CreateKeyVerb(VerbExtension):
...
         arg.completer = DirectoriesCompleter()
-        parser.add_argument('NAME', help='key name, aka ROS node name')
+        parser.add_argument(
+            '-c', '--security-context',
+            help='identity, aka ROS security context path')
 
     def main(self, *, args):
-        success = create_key(args.ROOT, args.NAME)
-        return 0 if success else 1
+        return 1 if create_key(args.keystore_root_path, args.security_context) is False else 0
```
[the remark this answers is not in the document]
ivanpauno: Solved in https://github.com/ros2/sros2/pull/177/commits/070105b15778030a10ff45a98d02964689752822.

sros2_cmake/README.md
ivanpauno: See https://github.com/ros2/sros2/pull/177/commits/070105b15778030a10ff45a98d02964689752822.

ivanpauno:
@mikaelarguedas @hidmic @kyrofa If there are not further comments, I will consolidate commits.

ivanpauno:
> Could you provide a repos file or a way of testing these changes?

You can try https://gist.githubusercontent.com/ivanpauno/6f6fdb864e0df355446b440bef1e4d71/raw/39e50c703887dbd4f7acbf2e784c98292c990443/ros2.repos.

sros2/sros2/verb/create_permission.py
```diff
@@ -31,17 +31,17 @@ class CreatePermissionVerb(VerbExtension):
     """Create permission."""
 
     def add_arguments(self, parser, cli_name):
-        arg = parser.add_argument('ROOT', help='root path of keystore')
+        arg = parser.add_argument('-k', '--keystore-root-path', help='root path of keystore')
```
kyrofa: This changes a required argument to an optional one, no? I don't see that actually being handled in the code. What happens if this isn't specified?

ivanpauno:
I've consolidated commits in the last force push.
I double checked the code was the same before pushing.

kyrofa:
I think we'd like to have a more thorough discussion of e96287bbee6467b11c2e69cb3bbbf8e437db264a , the CLI modifications. We don't want to hold this PR up though-- @ivanpauno how do you feel about removing it here?

ivanpauno:
> I think we'd like to have a more thorough discussion of [e96287b](https://github.com/ros2/sros2/commit/e96287bbee6467b11c2e69cb3bbbf8e437db264a) , the CLI modifications. We don't want to hold this PR up though-- @ivanpauno how do you feel about removing it here?

Sure.

If the only open discussion is https://github.com/ros2/sros2/pull/177#discussion_r401061155, making the arguments optional was definitely an error, so I can fix that.
In case we want to stick with positional arguments, I can revert that commit completely.

kyrofa:
Let's just revert for now. Not because we don't think the CLI needs to change, but because we want to get the context stuff shipped while giving the CLI the attention it deserves.

mikaelarguedas:
@ivanpauno I didnt get a chance to try the repos file locally. I looked at the CI you linked to and there seem to be quite a few security related tests failing (on arm and Windows). Is it expected?

ivanpauno:
> Let's just revert for now. Not because we don't think the CLI needs to change, but because we want to get the context stuff shipped while giving the CLI the attention it deserves.

Done!

ivanpauno:
> @ivanpauno I didnt get a chance to try the repos file locally. I looked at the CI you linked to and there seem to be quite a few security related tests failing (on arm and Windows). Is it expected?

Most of them are happening in `master` too, to be clearer:
- Linux CI is passing in `bionic`, and failing in `focal` because of https://github.com/ros2/sros2/issues/180.
- AArch64 failing tests are because I used `focal`, nightlies are showing the same failures.
- connext security tests have been failing on macOS for a long time, see https://github.com/ros2/build_farmer/issues/269. Same happened in the job I ran.
- I will investigate the ones introduced on Windows: https://ci.ros2.org/job/ci_windows/9674/#showFailuresLink.

sros2/setup.py
kyrofa: Can we add a reference to https://github.com/ros2/rclpy/issues/529 in this comment, please?

sros2/sros2/api/__init__.py
```diff
@@ -181,34 +223,35 @@ def create_keystore(keystore_path):
...
 def is_key_name_valid(name):
-    ns_and_name = name.rsplit('/', 1)
-    if len(ns_and_name) != 2:
-        print("The key name needs to start with '/'")
-        return False
-    node_ns = ns_and_name[0] if ns_and_name[0] else '/'
-    node_name = ns_and_name[1]
-
+    # TODO(ivanpauno): Use validate_security_context_name when it's propagated to `rclpy`.
+    #   This is not to bad for the moment.
```
kyrofa: Please add an issue reference here as well.

sros2/test/sros2/commands/security/verbs/test_create_key.py
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import datetime
-import os
+from pathlib import Path
```
kyrofa: Yay! Thank you for this.

sros2/test/sros2/commands/security/verbs/test_create_key.py
```diff
@@ -34,19 +34,20 @@
 
 # This fixture will run once for the entire module (as opposed to once per test)
 @pytest.fixture(scope='module')
-def node_keys_dir(tmpdir_factory):
-    keystore_dir = str(tmpdir_factory.mktemp('keystore'))
+def security_context_keys_dir(tmpdir_factory):
+    keystore_dir = Path(str(tmpdir_factory.mktemp('keystore')))
```
kyrofa: Has pytest been updated enough we can use `tmp_path_factory` here now? Would save you a little work.

sros2_cmake/package.xml
```diff
@@ -4,7 +4,7 @@
     <package format="3">
     <name>sros2_cmake</name>
     <version>0.8.1</version>
-    <description>CMake macros to configure security for nodes</description>
+    <description>Cmake macros to configure security</description>
```
kyrofa: Nit:

```suggestion
    <description>CMake macros to configure security</description>
```

sros2/test/sros2/commands/security/verbs/test_create_key.py
ivanpauno: Sure, but I think we can left that for a follow-up, because I'm not sure which `pytest` version we're using in the different platforms.

sros2_cmake/package.xml
ivanpauno: Done.

sros2/sros2/api/__init__.py
ivanpauno: Done.

kyrofa:
Alright this looks good to me now. What are we going to do about the failing tests, though? Can we ignore Focal for now? @ivanpauno let us know what you learn about Windows.

ivanpauno:
> Alright this looks good to me now. What are we going to do about the failing tests, though? Can we ignore Focal for now in order to get this in before the freeze? @ivanpauno let us know what you learn about Windows.

Yes, I think it's safe to ignore the tests that are failing in the nightlies.
Of course, there's a chance that I'm introducing a regression without realizing, but the chances are small if it's working ok in the other platforms.

ruffsl:
> https://github.com/ros2/sros2/pull/177#issuecomment-606794250

@ivanpauno if you revert/re-sign my commits here, could you move those reverted CLI commits to a new PR? I'd really prefer to remove any positional arguments in SROS from foxy before the API freeze. I think the PR only looks big because of the static xml files generated to test/track the xsd templates.

mikaelarguedas:
> I think the PR only looks big because of the static xml files generated to test/track the xsd templates.

The motivation for reverting it was not because of the size but the change but their nature. There were doubts on the relevance of making all these non-positional (discussion on matrix). Especially as as-is they were made optional. Currently `generate_artifacts` is using keyed optional args because they truly are all optional.  While the ones like `create_keystore` do need mandatory arguments and making them keyed doesn't serve much purpose while lengthening user invocation.

So in the end a more complete discussion need to be had and evaluate what UX we are trying to accomplish here. Discussion can happen on such PR though.

mikaelarguedas:
> Most of them are happening in master too, to be clearer:

Thanks @ivanpauno for the summary.
Figuring out the windows ones would be good indeed: there seem to be some fastrtps failures as well https://ci.ros2.org/job/ci_windows/9674/testReport/junit/(root)/projectroot/test_security_nodes_c__rmw_fastrtps_cpp_2/

Is there a way for maintainers here want to be notified for sros2/test_security failing jobs? I see some of those have been failing for a long time but we weren't aware of it.

ivanpauno:
> Alright this looks good to me now. What are we going to do about the failing tests, though? Can we ignore Focal for now in order to get this in before the freeze?

I will ignore all existing failures on nightlies. Unfortunately, there's a minimal chance I introduce a regression by doing this.

> @ivanpauno if you revert/re-sign my commits here, could you move those reverted CLI commits to a new PR? I'd really prefer to remove any positional arguments in SROS from foxy before the API freeze. I think the PR only looks big because of the static xml files generated to test/track the xsd templates.

https://github.com/ros2/sros2/tree/contexts-backup has the original commits of the PR.

> @ivanpauno let us know what you learn about Windows.

That's the last failure I see in my jobs I that's not on the nightlies.
If I'm lucky, I will find the reason today.

> Is there a way for maintainers here want to be notified for sros2/test_security failing jobs? I see some of those have been failing for a long time but we weren't aware of it.

IDK. @ros2/team for feedback.

jacobperron:
> Is there a way for maintainers here want to be notified for sros2/test_security failing jobs? I see some of those have been failing for a long time but we weren't aware of it.

I think at least the ROS 2 buildfarmer should open issues for failures (e.g. on sros2) and consider tagging @ros-security-wg if the issue is in a different repository.

ivanpauno:
> Figuring out the windows ones would be good indeed: there seem to be some fastrtps failures as well https://ci.ros2.org/job/ci_windows/9674/testReport/junit/(root)/projectroot/test_security_nodes_c__rmw_fastrtps_cpp_2/

The windows failures of security tests were those, and the same test was failing with connext. Fixed them in https://github.com/ros2/system_tests/pull/406/commits/1d93d97e105eb049b955ccfef49a1b5ef32e822d.

I will re-run CI today by end of day. If everything looks good tomorrow, I will merge.

ivanpauno:
> lgtm, a bionic CI on aarch64 would be good to confirm this doesn't break anything on that platform

Done. We're not supporting `bionic` for Foxy, but running it will add extra confidence in that `sros2` wasn't broken.

These are the last jobs I ran https://github.com/ros2/rmw_fastrtps/pull/312#issuecomment-607504858 (only with unrelated failures), and I have to wait to these two before merging https://github.com/ros2/rmw_fastrtps/pull/312#issuecomment-607829112.

Hopefully, I will be merging today.

sros2/test/policies/permissions/add_two_ints/permissions.xml
```diff
@@ -0,0 +1,106 @@
+<dds xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://www.omg.org/spec/DDS-SECURITY/20170901/omg_shared_ca_permissions.xsd">
+  <permissions>
+    <grant name="/add_two_ints/add_two_ints_server">
+      <subject_name>CN=/add_two_ints/add_two_ints_server</subject_name>
+      <validity>
+        <not_before>2013-10-26T00:00:00</not_before>
+        <not_after>2023-10-26T22:45:30</not_after>
```
hidmic, line 7: @ruffsl @ivanpauno meta: is this supposed to span the foreseeable lifetime of Foxy? Won't this bite us in 3 years time? Same elsewhere.

mikaelarguedas: this is something I'm planning to address separately as unrelated to this PR.

This should be using the time of generation as the start date similarly to how it's done for the certificates. I'll open an issue to make sure it doesnt slip off.
Hackier way to address this is to just change the hardcoded value to start in 2020 which will give us a lot of time to implement it more properly ^^

sros2/sros2/api/__init__.py
```diff
@@ -181,34 +223,35 @@ def create_keystore(keystore_path):
...
     try:
-        return (validate_namespace(node_ns) and validate_node_name(node_name))
-    except (InvalidNamespaceException, InvalidNodeNameException) as e:
-        print('{}'.format(e))
+        return validate_namespace(name)
+    except InvalidNamespaceException as e:
+        print(f'{e}')
```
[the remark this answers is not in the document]
ivanpauno: Fixed.
I will avoid re-running CI because of this, as it's just an style change.

sros2/test/policies/permissions/add_two_ints/permissions.xml
hidmic: I'd rather address #186 instead of pushing the expiration date (and then forgetting about it). Thanks for the explanation @mikaelarguedas!

mikaelarguedas: Me too. Happy to review any PR addressing #186, timing before feature freeze is pretty tight so if it falls on my plate I'll most likely just push the date back

PR #178, opened 2020-03-16
Add GitHub actions for linting and source-build CI

mikaelarguedas:
Currently Ubuntu only and no xmllint test for `sros2` due to schema hosting issue.

Looking for feedback on the following:
- the test workflow runs daily
- the lint workflow runs on push and PR events only
- log upload: usefulness of uploading colcon logs?
  Some repositories using the ROS github actions upload the colcon logs (e.g. [rosbag2](https://github.com/ros2/rosbag2/blob/f3228aa784ea15066ede01898f009a0d2de47c34/.github/workflows/test.yml#L20)) but they upload them only on job success, which doesn't seem very useful. I opted to not add that upload step until we come across cases where we need the logs.
 
Examples of jobs using these parameters:
Build + test: https://github.com/mikaelarguedas/sros2/actions/runs/56872161
Lint: https://github.com/mikaelarguedas/sros2/actions/runs/56872157

---

Failing on Macos for now, haven't tried windows:

<details>

```
E   RuntimeError: Failed to initialize init_options: failed to find shared library of rmw implementation. Searched rmw_fastrtps_cpp, at /Users/runner/runners/2.165.2/work/sros2/sros2/ros_ws/src/ros2/rmw_implementation/rmw_implementation/src/functions.cpp:61, at /Users/runner/runners/2.165.2/work/sros2/sros2/ros_ws/src/ros2/rcl/rcl/src/rcl/init_options.c:55
```

</details>

Possible reason: environment may be sanitized (maybe SIP is enabled) so the relevant libs are not on the `DYLD_LIBRARY_PATH`: upstream ticket  https://github.com/ros-tooling/action-ros-ci/issues/81

@kyrofa did you ever succeed to run some ROS nodes on MacOS using this action ?

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,2 @@
+fixes:
+  - "ros_ws/src/sros2/::"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,9 +47,9 @@ jobs:
         apt-get -qq update
         apt-get -qq upgrade -y
         rosdep update
-        rosdep install -y --from-paths . --ignore-src --rosdistro foxy
-    - run: . /opt/ros/foxy/setup.sh && colcon build
-    - run: . /opt/ros/foxy/setup.sh && colcon test
+        rosdep install -y --from-paths . --ignore-src --rosdistro $ROS_DISTRO
+    - run: . /opt/ros/$ROS_DISTRO/setup.sh && colcon build
+    - run: . /opt/ros/$ROS_DISTRO/setup.sh && colcon test
     - run: colcon test-result
     - name: Upload Logs
       uses: actions/upload-artifact@v1
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,7 +49,7 @@ jobs:
         rosdep update
         rosdep install -y --from-paths . --ignore-src --rosdistro $ROS_DISTRO
     - run: . /opt/ros/$ROS_DISTRO/setup.sh && colcon build
-    - run: . /opt/ros/$ROS_DISTRO/setup.sh && colcon test
+    - run: . /opt/ros/$ROS_DISTRO/setup.sh && colcon test --executor sequential --event-handlers console_direct+
     - run: colcon test-result
     - name: Upload Logs
       uses: actions/upload-artifact@v1
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,7 +4,7 @@ on:
   push:
   schedule:
     # Run daily
-    - cron:  '0 12 * * *'
+    - cron:  '0 20 * * *'
 
 jobs:
   test_latest:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,8 +20,9 @@ jobs:
         package-name: |
           sros2
           sros2_cmake
-          # commented for now https://github.com/eProsima/Fast-RTPS/issues/1087
-          # test_security
+        # skipping end-to-end tests for now https://github.com/eProsima/Fast-RTPS/issues/1087
+        #  test_security
+        # extra-cmake-args: '-DSECURITY=ON --no-warn-unused-cli'
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v1.0.6
       if: github.event_name == 'pull_request' || (github.event_name == 'push' && github.ref == 'refs/heads/master')
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,56 @@
+name: Test sros2
+on:
+  pull_request:
+  push:
+  schedule:
+    # Run daily
+    - cron:  '0 12 * * *'
+
+jobs:
+  test_latest:
+    runs-on: ubuntu-latest
+    container: osrf/ros2:devel
+    steps:
+    - run: |
+        apt-get -qq update
+        apt-get -qq upgrade -y
+        apt-get -qq install -y curl libasio-dev libtinyxml2-dev
+    - uses: mikaelarguedas/action-ros-ci@sros2-version
+      with:
+        package-name: |
+          sros2
+          sros2_cmake
+    - name: Upload coverage to Codecov
+      uses: codecov/codecov-action@v1.0.6
+      if: github.event_name == 'pull_request' || (github.event_name == 'push' && github.ref == 'refs/heads/master')
+      with:
+        file: ros_ws/build/sros2/coverage.xml
+        flags: unittests
+        name: sros2-coverage
+        fail_ci_if_error: true
+        yml: .codecov.yml
+    - name: Upload Logs
+      uses: actions/upload-artifact@v1
+      if: failure()
+      with:
+        name: colcon-logs
+        path: ros_ws/log
+  test_nightly:
+    runs-on: ubuntu-latest
+    container: osrf/ros2:nightly
+    steps:
+    - uses: actions/checkout@v1
+    - run: |
+        apt-get -qq update
+        apt-get -qq upgrade -y
+        rosdep update
+        rosdep install -y --from-paths . --ignore-src --rosdistro foxy
+    - run: . /opt/ros/foxy/setup.sh && colcon build
+    - run: . /opt/ros/foxy/setup.sh && colcon test
+    - run: colcon test-result
+    - name: Upload Logs
+      uses: actions/upload-artifact@v1
+      if: failure()
+      with:
+        name: colcon-logs
+        path: ros_ws/log
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,6 +15,8 @@ jobs:
         apt-get -qq update
         apt-get -qq upgrade -y
         apt-get -qq install -y curl libasio-dev libtinyxml2-dev
+    # TODO(mikaelarguedas) switch back to ros-tooling/action-ros-ci once
+    # https://github.com/ros-tooling/action-ros-ci/pull/109 is released
     - uses: mikaelarguedas/action-ros-ci@sros2-version
       with:
         package-name: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,4 +1,4 @@
-name: Test sros2
+name: SROS2 CI
 on:
   pull_request:
   push:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,6 +20,8 @@ jobs:
         package-name: |
           sros2
           sros2_cmake
+          # commented for now https://github.com/eProsima/Fast-RTPS/issues/1087
+          # test_security
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v1.0.6
       if: github.event_name == 'pull_request' || (github.event_name == 'push' && github.ref == 'refs/heads/master')
--- a/README.md
+++ b/README.md
@@ -1,5 +1,8 @@
 # Context
 
+![SROS2 CI](https://github.com/ros2/sros2/workflows/SROS2%20CI/badge.svg)
+[![codecov](https://codecov.io/gh/ros2/sros2/branch/master/graph/badge.svg)](https://codecov.io/gh/ros2/sros2)
+
 This package provides the tools and instructions to use ROS2 on top of DDS-Security.
 The security feature is tested across platforms (Linux, macOS, and Windows) as well as across different languages (C++ and Python).
 
--- a/sros2/.coveragerc
+++ b/sros2/.coveragerc
@@ -1,0 +1,4 @@
+[run]
+omit =
+    # omit test directory
+    test/*


kyrofa:
MacOS and Windows have been a bit flaky. We're still running them in nodl, and MacOS passes, but Windows consistently fails due to Fast RTPS not building. We only require Linux to pass given the flakiness.

mikaelarguedas:
> We're still running them in nodl, and MacOS passes

Oh nice. Do your tests run any ROS Nodes or are they just pure python tests not relying on a running ROS system ?

kyrofa:
I believe they're just pure Python at the moment, which probably explains the difference.

mikaelarguedas:
This is now uploading colcon logs only on failure.

build and test: https://github.com/mikaelarguedas/sros2/actions/runs/58114846
lint: https://github.com/mikaelarguedas/sros2/actions/runs/58114842

kyrofa:
Note that having separate test files means you also have separate pass/fail badges:

Linter: Lint sros2 repo
Tests: Test sros2

I kinda prefer just having one, but what do you think?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,62 @@
...
+    name: ament_lint_cmake
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - uses: ros-tooling/setup-ros@0.0.13
+    - uses: ros-tooling/action-ros-lint@0.0.5
+      with:
+        linter: copyright
+        package-name: |
+          sros2_cmake
```
kyrofa: Aren't you linting `sros2_cmake` in the above job?

mikaelarguedas:
>I kinda prefer just having one, but what do you think?

That's a good point.

I default to multiple files because one issue I have with single workflow is that it means same triggers for all jobs in the workflow. You can do some conditionals on triggers but it quickly becomes limited.This is why I use different ones in e.g. https://github.com/osrf/docker_images#table-of-contents.


In this case it doesn't make much of a difference, merging them into the same workflow means triggering linter jobs along the nightly builds and test, which is totally fine.

.github/workflows/test.yml
```diff
@@ -0,0 +1,27 @@
...
+  schedule:
+    # Run daily
+    - cron:  '0 12 * * *'
+
+jobs:
+  build_and_test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
```
kyrofa: Given the ROS 2 release strategy, should this actually be `ubuntu-18.04`? This question applies to the linters, as well.

.github/workflows/test.yml
```diff
@@ -0,0 +1,27 @@
+name: Test sros2
+on:
+  pull_request:
+  push:
+  schedule:
+    # Run daily
+    - cron:  '0 12 * * *'
```
kyrofa: Also, I have found this to be of limited usefulness, as we discussed. No strong opinions though.

kyrofa:
> I default to multiple files because one issue I have with single workflow is that it means same triggers for all jobs in the workflow.

Good point, that makes sense.

> In this case it doesn't make much of a difference, merging them into the same workflow means triggering linter jobs along the nightly builds and test, which is totally fine.

Sounds like neither of us feel too strongly about this. Let's ask this, then: do we want to show that badge somewhere? In the README, maybe? Do we care about them at all? If we do, let's make it one job. If we don't, let's leave it as two.

.github/workflows/lint.yml
mikaelarguedas: indeed, a copy-n-paste mistake, it should be `lint_cmake`, let me see if I can find a way to not have to list the linter name twice for every job.

Edit: actually the current lint action doesn't support linters with underscore/dashes so I guess we'll just remove this one for now

.github/workflows/test.yml
mikaelarguedas: As this is meant to be the day-to-day development I was thinking that latest was better so that we can switch to 20.04 automatically.
Granted IDK when Github Actions will provide a more recent runner while ROS2 CI should migrate shortly. Whichraises the question of how valid the ros-ci-action approach will be if Github doesnt provide a 20.04 environment in the near future.

I'll switch to ubuntu-18-04 for now and move forward with the actions approach and we can revisit in a couple months.

.github/workflows/test.yml
mikaelarguedas: on some leaf projects this actually helped us keep up with API and behavior changes.

Arguably this is part of `ros2.repos` so I guess there are less surprises of API changes needing to be cought by this CI.

I'll keep it so that we (not sure who "we" is in this context) receive emails automatically if it starts failing

.github/workflows/lint.yml
gavanderhoorn: As this is YAML, you should be able to use anchors and aliases.

But, [this](https://github.community/t5/GitHub-Actions/Support-for-YAML-anchors/td-p/30336) seems to suggest it's not supported in GH actions.

mikaelarguedas:
> Sounds like neither of us feel too strongly about this. Let's ask this, then: do we want to show that badge somewhere? In the README, maybe? Do we care about them at all? If we do, let's make it one job so we only have one to show. If we don't, let's leave it as two..

Sounds good to me, it's now in the README.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,94 @@
...
+
+jobs:
+  build_and_test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04]
+    steps:
+    - uses: ros-tooling/setup-ros@0.0.13
+    - uses: ros-tooling/action-ros-ci@0.0.13
```
mikaelarguedas: RfC: not sure how to manage action versions in workflows. Is there a way to bump those automatically? otherwise maybe we should just point at the `master` branch instead of a specific version?

.github/workflows/lint.yml
mikaelarguedas: Yeah hopefully it'll happen in the near future, other systems like travis and Circle do support it and it is pretty convenient.

Ended up using some matrix env hackery instead... https://github.com/ros2/sros2/pull/178/files#diff-e9f950f17198d3d5e3122a44230a09b9R31
not great but allows some deduplication.

.github/workflows/ci.yml
kyrofa: If I remember correctly you can leave off the version entirely, but I think it's probably smart to leave it pinned and update consciously or things will break on us and require a flag day to fix.

.github/workflows/test.yml
kyrofa: Very good, let's leave it then.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,94 @@
...
+          sros2_cmake
+    - name: Upload Logs
+      uses: actions/upload-artifact@v1
+      if: failure()
+      with:
+        name: colcon-logs
+        path: ros_ws/log
+
+  # Linter jobs
+  ament_copyright:
```
kyrofa: Re-thinking this, is it actually necessary to have this given that we also have [this](https://github.com/ros2/sros2/blob/master/sros2/test/test_copyright.py)? Doesn't it mean that we're running it as part of the test job anyway?

.github/workflows/ci.yml
artivis: Pointing directly to `master` opens the way for breakage, I think we better manually bump it as we see fit.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,94 @@
...
+      matrix:
+        os: [ubuntu-18.04]
+    steps:
+    - uses: ros-tooling/setup-ros@0.0.13
+    - uses: ros-tooling/action-ros-ci@0.0.13
+      with:
+        package-name: |
+          sros2
+          sros2_cmake
+    - name: Upload Logs
```
artivis: Total noob questions, where are logs uploaded? How to access them? Who can access them?

mikaelarguedas: > Total noob questions

We're all discovering here :)

>  where are logs uploaded?

They will be uploaded on that job run. So going to that run's page on github should give you access to the artifacts. e.g. https://github.com/mikaelarguedas/sros2/actions/runs/58398774

> Who can access them?

Everyone I believe

.github/workflows/ci.yml
mikaelarguedas: >  Doesn't it mean that we're running it as part of the test job anyway?

Yes we are

The main advantage I see for this is a faster feedback loop, these linter jobs will give result ~10times faster than the test job or the buildfarm.
What seems to be the most commonly overlooked parts in contributions are usually style (and coverage). Being able to give fast feedback on those can help reduce iteration loop time and the number of iterations. 

So I tend towards keeping those for now and consider removing them if it proves to not be helpful or give results not matching the ones from the buildfarm

.github/workflows/ci.yml
mikaelarguedas: sounds good, kept the pinned approach but bumped to latest release at https://github.com/ros2/sros2/pull/178/commits/3cb4573242b9ad304e69918591145369389f3ffe

ros-discourse:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2020-03-18/13313/1

.github/workflows/ci.yml
kyrofa: Fair point, as long as we realize:

- We're running some of them twice (once in the linter job, once in the test job)
- We're running a different set of linters in the build farm due to the fact that we don't have colcon tests for all of the ones here

.github/workflows/ci.yml
```diff
@@ -0,0 +1,94 @@
...
+    - uses: actions/checkout@v1
+    - uses: ros-tooling/setup-ros@0.0.15
+    - uses: ros-tooling/action-ros-lint@0.0.5
+      with:
+        linter: ${{ matrix.linter_name }}
+        package-name: |
+          sros2
+          sros2_cmake
+
+  # action-ros-lint does not support multi work linters like lint_cmake
```
kyrofa: Are we confident this will change soon-ish? Or should we remove this section so it doesn't turn into crufty comments?

emersonknapp: I'd say remove it, I don't personally believe in checking in commented-out code. I don't have a lot of context on this but I did notice that there wasn't a feature request or bug report opened up for it. Please do not hesitate in the future @mikaelarguedas ! I opened this one https://github.com/ros-tooling/action-ros-lint/issues/31

kyrofa: Ah, thanks @emersonknapp!

mikaelarguedas: > Please do not hesitate in the future @mikaelarguedas ! I opened this one ros-tooling/action-ros-lint#31

I held on as it was a 2 line-fix, I was planning on contributing it rather that opening a feature request ;)
I now had the opportunity to test it so opened the PR is at https://github.com/ros-tooling/action-ros-lint/pull/34

mikaelarguedas:
I left this PR hanging the last couple days as ros2 builfarm switched to 20.04 over the weekend. This puts a bigger question mark on the relevance of this set of actions.
I'm exploring alternatives  e.g. https://github.com/mikaelarguedas/sros2/actions/runs/61362217 that allow to run faster and match closely the buildfarm because based on the nightly builds.

Leaving this one on hold for now

mikaelarguedas:
New version of this PR:

Workflow with 2 jobs:
- job building on top of `osrf/ros2:nightly`, provides quick feedback for failures such as linters (~2min)
- job building from source on top of `osrf/ros2:devel`, takes longer and provides more accurate representation of the code if PR was merged now (~20min)
  - provides coverage information with feedback on the PR
  - runs integration tests (`test_security` package, though disabled for now as they are not passing on Focal)

Example of repo/PR using this: https://github.com/mikaelarguedas/sros2/pull/2



Prerequisites for merging:
- enable codecov.io for this repository

Related work:
- coverage:
  - mixin https://github.com/colcon/colcon-mixin-repository/pull/21
  - colcon default reporting: https://github.com/colcon/colcon-core/pull/331
  - action-ros-ci: https://github.com/ros-tooling/action-ros-ci/pull/109

---

Future work:
There is still custom logic in this repo. Ideally we could modify the ros-tooling actions to meet our need and limit custom code to a minimum, some examples of improvements

- switch back to use upstream action-ros-ci
- improve action-ros-ci (non exhaustive):
  - standalone: action should install its own required dependencies
  - more control to users: not hardcoding colcon arguments or allow users to override any argument
  - consistent failure modes (`colcon lcov-result --initial` fails but doesn't fail the build but `colcon lcov-result` does o_0)
  - allow to specify rosdep arguments
  - mode using current build dir and not pulling any repos file
  - print test result summary
- cache docker images daily to avoid pulling (80% of the time of the nightly job)

sros2/.coveragerc
```diff
@@ -0,0 +1,4 @@
+[run]
+omit =
+    # omit test directory
+    test/*
```
mikaelarguedas: I did this to get accurate coverage numbers. Code in test is usually pretty well covered and skewing the numbers

.codecov.yml
```diff
@@ -0,0 +1,2 @@
+fixes:
+  - "ros_ws/src/sros2/::"
```
mikaelarguedas: This rewrites the path of the files in the report so that we can browse them directly on the codecov.io interface and see which lines/conditionals are covered and which are not

.github/workflows/test.yml
```diff
@@ -0,0 +1,59 @@
...
+jobs:
+  test_latest:
+    runs-on: ubuntu-latest
+    container: osrf/ros2:devel
+    steps:
+    - run: |
+        apt-get -qq update
+        apt-get -qq upgrade -y
+        apt-get -qq install -y curl libasio-dev libtinyxml2-dev
+    - uses: mikaelarguedas/action-ros-ci@sros2-version
```
mikaelarguedas: Here we are using a custom version to:
- use colcon provided argument for pytest coverage instead of overriding them. More details at  https://github.com/ros-tooling/action-ros-ci/pull/109
- Avoid having to install pip and the unreleases colcon lcov-result that otherwise if making ci fail https://github.com/ros-tooling/action-ros-ci/commit/f9c55f383bd617b8b39e7346124db1d2b9b9922b

.github/workflows/test.yml
```diff
@@ -0,0 +1,59 @@
...
+  pull_request:
+  push:
+  schedule:
+    # Run daily
+    - cron:  '0 20 * * *'
+
+jobs:
+  test_latest:
+    runs-on: ubuntu-latest
+    container: osrf/ros2:devel
```
kyrofa: What does this actually mean? `runs-on` is the VM used by the action, but then we're saying "please also run the rest of this in the given docker container"?

.github/workflows/test.yml
```diff
@@ -0,0 +1,59 @@
...
+        # extra-cmake-args: '-DSECURITY=ON --no-warn-unused-cli'
+    - name: Upload coverage to Codecov
+      uses: codecov/codecov-action@v1.0.6
+      if: github.event_name == 'pull_request' || (github.event_name == 'push' && github.ref == 'refs/heads/master')
+      with:
+        file: ros_ws/build/sros2/coverage.xml
+        flags: unittests
+        name: sros2-coverage
+        fail_ci_if_error: true
+        yml: .codecov.yml
```
kyrofa: Do we have the ability to fail CI if coverage decreases?

.github/workflows/test.yml
```diff
@@ -0,0 +1,59 @@
...
+        path: ros_ws/log
+  test_nightly:
+    runs-on: ubuntu-latest
+    container: osrf/ros2:nightly
+    steps:
+    - uses: actions/checkout@v1
+    - run: |
+        apt-get -qq update
+        apt-get -qq upgrade -y
+        rosdep update
```
kyrofa: No init needed?

.github/workflows/test.yml
kyrofa: Given that it sounds like these issues will be fixed upstream, would you mind making a comment similar to this so we know in the future 1) why we're using a custom version and 2) when it should be revisited?

.github/workflows/test.yml
mikaelarguedas: Not at the moment. And it seems hard to implement outside of the codecov machinery as the job itself doesn't know past coverage, only the coverage status of the current run. Past values are stored on codecov.io and the comparison and feedback is handled on their side.
However codecov will post a comment on the PR with the coverage diff (e.g. https://github.com/mikaelarguedas/sros2/pull/2#issuecomment-604699896) and that will give both contributor and maintainer notification of coverage going up or down on the PR.

.github/workflows/test.yml
mikaelarguedas: In short: all `ros:*` and `osrf/ros*` images install and init rosdep for you, so we should only ever `update`.


`init` is (or IMO should always be) done when the `rosdep` package is installed. As init cannot be ran twice without error, I believe users should not get in the habit of running it in scripts where they don't also handle the installation of rosdep.

.github/workflows/test.yml
mikaelarguedas: That is my understanding.

`runs-on` is a required element so the workflow file parsing will fail if it's not set
If it's set to anything other than ubuntu, the job will fail because github actions dont support docker on other platforms (hopefully that will change in the future and support windows containers)

.github/workflows/test.yml
kyrofa: Alright fair enough.

.github/workflows/test.yml
kyrofa: Other than this comment, this looks good to me.

.github/workflows/test.yml
mikaelarguedas: To clarify this a bit further, the github actions job will not fail, but codecov does post a comment *and* status checks on the PR (codecov/project and codecov/patch). If the patch decreases coverage, the codecov status check will be a failed one. Valid example at the time of the writing https://github.com/ros-infrastructure/rosdep/pull/712

.github/workflows/test.yml
mikaelarguedas: added comment at https://github.com/ros2/sros2/pull/178/commits/535a700bb73a831bfb70e0020e5f3a31f44472b6

Arnatious:
LGTM

artivis:
About the prerequisite
> enable codecov.io for this repository

Who's responsibility is this? I don't know `codecov.io` very well, can we create groups or whatnot so that the admin work is shared?

kyrofa:
> Who's responsibility is this?

Mine, this is already done.

artivis:
> Mine, this is already done.

Ok, good to know.
We'll have to make sure that this can be managed by other WG members in case it is necessary.

kyrofa:
It was actually already enabled and (at least we think) actions don't need a token, so there's really nothing to do there.

README.md
```diff
@@ -1,5 +1,8 @@
 # Context
 
+![SROS2 CI](https://github.com/ros2/sros2/workflows/SROS2%20CI/badge.svg)
+[![codecov](https://codecov.io/gh/ros2/sros2/branch/master/graph/badge.svg)](https://codecov.io/gh/ros2/sros2)
```
artivis: `SROS2 CI` badge does not currently render. I guess that will be fixed after merge and a first CI run?

kyrofa: Indeed. Check [this](https://github.com/mikaelarguedas/sros2/workflows/SROS2%20CI/badge.svg) out to see the current state of it.

artivis:
Left a last small comment/question.
LGTM :+1:

kyrofa:
Alright, tests for sros2:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9865/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5481/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8025/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9721/)

mikaelarguedas:
> Mine, this is already done. We think stuck_out_tongue .

So apparently it was not done ^^
it does upload reports to codecov, but it does not process and post back progress on the PRs.

2 ways to do it:
- integrate the codecov github app for ros2/sros2
- add a webhook on ros2/sros2 for codecov to use

App integration seems the way to go these days but may require permission at the org level asd the app doesnt seem to be installed on the ros2 org.

Is there a preference from @ros2/team or @ros2/security-members for one approach or the other ?

kyrofa:
Huh, maybe that's why it looked like it was already enabled, the default settings didn't need the app installed. I prefer that, of course, but it's up to the org admins.

PR #188, opened 2020-04-03
api: reorganize keystore API

kyrofa:
We don't currently have a nicely controlled/organized API: everything is implemented in the api's `__init__.py`, making is way longer than it needs to be, and most is public without needing to be, which makes it hard to change once we release. This PR starts the process of resolving that by extracting the keystore API into its own succinct, non-public module. In the end we'll hopefully have a tight, small public API to maintain.

### DIFF
--- a/sros2/sros2/api/__init__.py
+++ b/sros2/sros2/api/__init__.py
@@ -13,17 +13,13 @@
 # limitations under the License.
 
 from collections import namedtuple
-import datetime
 import errno
 import os
 import sys
 
 from cryptography import x509
 from cryptography.hazmat.backends import default_backend as cryptography_backend
-from cryptography.hazmat.bindings.openssl.binding import Binding as SSLBinding
-from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives import serialization
-from cryptography.hazmat.primitives.asymmetric import ec
 
 from lxml import etree
 
@@ -33,23 +29,17 @@ from rclpy.validate_namespace import validate_namespace
 
 from sros2.policy import (
     get_policy_default,
-    get_transport_default,
     get_transport_schema,
     get_transport_template,
     load_policy,
 )
 
-HIDDEN_NODE_PREFIX = '_'
-DOMAIN_ID_ENV = 'ROS_DOMAIN_ID'
+from . import _keystore, _utilities
 
-_DEFAULT_COMMON_NAME = 'sros2testCA'
+HIDDEN_NODE_PREFIX = '_'
 
 NodeName = namedtuple('NodeName', ('node', 'ns', 'fqn'))
 TopicInfo = namedtuple('Topic', ('fqn', 'type'))
-
-KS_CONTEXT = 'contexts'
-KS_PUBLIC = 'public'
-KS_PRIVATE = 'private'
 
 RMW_WITH_ROS_GRAPH_INFO_TOPIC = ('rmw_fastrtps_cpp', 'rmw_fastrtps_dynamic_cpp')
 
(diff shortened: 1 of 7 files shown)


kyrofa:
Hmm, having some issues uploading the coverage it seems...

```
Error: connect ETIMEDOUT 35.199.43.247:443
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1129:14) {
  errno: 'ETIMEDOUT',
  code: 'ETIMEDOUT',
  syscall: 'connect',
  address: '35.199.43.247',
  port: 443
}
```

sros2/sros2/api/__init__.py
```diff
@@ -291,48 +144,53 @@ def create_permission(keystore_path, identity, policy_file_path):
 
 
 def create_permissions_from_policy_element(keystore_path, identity, policy_element):
-    domain_id = os.getenv(DOMAIN_ID_ENV, '0')
     relative_path = os.path.normpath(identity.lstrip('/'))
-    key_dir = os.path.join(keystore_path, KS_CONTEXT, relative_path)
+    key_dir = os.path.join(_keystore.keystore_context_dir(keystore_path), relative_path)
```
Arnatious, line 148: Since we're refactoring everything anyway, would it make sense to sub in pathlib for readability?

kyrofa: That's the thing though-- we're not actually refactoring anything here. That would require a more in-depth review and this simple re-organization can already be a pretty significant diff to look through. As it stands I can promise that no meaningful changes were made, which makes review a bit faster. Once this re-organization is finished, we can change to using pathlib without breaking API, so it isn't time-sensitive.

mikaelarguedas:
> I'll try re-running in a bit. Update: Aw heck, now the other one is failing to check the code out. Maybe github is having issues at the moment.

These seem to be more of a github actions issue than a github outage issue...
A hidden issue that is happening (coverage still not being uploaded..) is that github actions doesnt populate the pull request name properly on reruns... https://github.com/actions/checkout/issues/58#issuecomment-589447479

hopefully most of these issues will be fixes by https://github.com/ros2/sros2/pull/189 :crossed_fingers: 

Otherwise just amending you commig would make this a new workflow run and work around the issues related to "rerunning a failed workflow"

sros2/sros2/api/_utilities.py
```diff
@@ -0,0 +1,154 @@
+# Copyright 2020 Canonical Ltd
```
mikaelarguedas, line 1: This should retain the original copyright notice. The code in this file predates 2020 and was not written solely by Canonical.

Similarly history should be preserved

Arnatious:
@kyrofa per @mikaelarguedas request https://stackoverflow.com/questions/38729453/how-to-preserve-git-history-when-refactoring-into-multiple-files

sros2/sros2/api/_utilities.py
kyrofa: Good catch, thank you. I also redid this to keep history, but I suspect that means this cannot be squashed.

kyrofa:
> General question: the new modules are prefixed with _, does that make them private?

Yes, although given that nothing is truly private in python, these are simply "non-public". It's simply a way of telling users what we're promising to maintain through a release and what might change.

> In these (private?) modules some functions are private and some public, what does that make them ? all private?

It's the same public/non-public idea applied to the module. Underscore-prefixed things are intended for use within the module. Ideally such things would be able to change without needing to change any callers (other than the occasional test). The non-underscore-prefixed functions are intended for use outside of the module (e.g. by verbs). We could be using `__all__` as well, although I'm not sure it gains us much.

Anyway, in this PR the modified verbs are now using non-public API. That's okay because they're in the same package, and even the non-public modules have their own public/non-public API nicely defined. This comes in handy when it comes to external packages using the SROS2 API: this PR makes the keystore stuff all non-public, because the `api` package no longer contains any public keystore-related stuff. That's not necessarily the end goal, but this series of PRs will refactor it all to be non-public and we can then chat about what we actually want to support as public API for the LTS. Sound okay?

mikaelarguedas:
> this PR makes the keystore stuff all non-public, because the api package no longer contains any public keystore-related stuff. That's not necessarily the end goal, but this series of PRs will refactor it all to be non-public and we can then chat about what we actually want to support as public API for the LTS (probably by importing it back into __init__.py). Sound okay?

Sounds good to me :+1:

---

Looks like the splitting with history didn't succeed, if I try to git blame the file I dont see the commits before the split: https://github.com/ros2/sros2/blame/300435a668e02bfa0118a2302621d169799afe06/sros2/sros2/api/_keystore.py

I was more expecting something like https://github.com/mikaelarguedas/sros2/blame/split-with-history/sros2/sros2/api/_keystore.py.

I can push a version with fixed history on this branch if you want

kyrofa:
> Looks like the splitting with history didn't succeed

Dang that was a lot harder than it should have been, but I think I finally got it. Blames look [way better](https://github.com/ros2/sros2/blame/064fb08c52b206a789300c16a8f0799b01c4297f/sros2/sros2/api/_keystore.py) now.

sros2/sros2/api/_utilities.py
```diff
@@ -0,0 +1,155 @@
...
+from cryptography.hazmat.backends import default_backend as cryptography_backend
+from cryptography.hazmat.bindings.openssl.binding import Binding as SSLBinding
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives import serialization
+from cryptography.hazmat.primitives.asymmetric import ec
+
+_DOMAIN_ID_ENV = 'ROS_DOMAIN_ID'
+
+
+def create_symlink(*, src, dst):
```
artivis, line 29: This was initially 'non-public'. How about keeping it so in order to possibly move it to a more generic place later on?

sros2/sros2/api/_keystore.py
```diff
@@ -0,0 +1,150 @@
...
+from . import _utilities
+
+
+_KS_CONTEXT = 'contexts'
+_KS_PUBLIC = 'public'
+_KS_PRIVATE = 'private'
+_DEFAULT_COMMON_NAME = 'sros2testCA'
+
+
+def create_keystore(keystore_path):
```
artivis, line 33: *Outside the scope of this PR*:
Some of those functions wording are now unfortunate given that they live in `_keystore`, e.g. `_keystore.create_keystore`.

sros2/sros2/api/_keystore.py
```diff
@@ -0,0 +1,150 @@
...
+    return (
+        os.path.isfile(os.path.join(path, _KS_PUBLIC, 'permissions_ca.cert.pem')) and
+        os.path.isfile(os.path.join(path, _KS_PUBLIC, 'identity_ca.cert.pem')) and
+        os.path.isfile(os.path.join(path, _KS_PRIVATE, 'permissions_ca.key.pem')) and
+        os.path.isfile(os.path.join(path, _KS_PRIVATE, 'identity_ca.key.pem')) and
+        os.path.isfile(os.path.join(path, _KS_CONTEXT, 'governance.p7s'))
+    )
+
+
+def keystore_context_dir(keystore_path: str) -> str:
```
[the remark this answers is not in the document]
kyrofa: Alright, `get_` has been added to all three.

kyrofa:
Hmm, something broke here again @mikaelarguedas, any ideas?

```
CMake Error at /usr/share/cmake-3.16/Modules/FindPackageHandleStandardArgs.cmake:146 (message):
    Could NOT find PythonLibs (missing: PYTHON_LIBRARIES PYTHON_INCLUDE_DIRS)
    (Required is at least version "3.5")
```

mikaelarguedas:
> Hmm, something broke here again @mikaelarguedas, any ideas?

Yes this is the same breakage as the daily sros2 builds that we mentioned on matrix. Issue is outside this repo, a workaround has been pushed and this is now fixed

TL;DR:
This is a combination of a change in the docker images, a bug of rosdep and some unexpected side effect of the new release of rosdep that all happened in the same 24 hours.
docker_images fix for new release of rosdep https://github.com/osrf/docker_templates/pull/85
rosdep bug tracked here https://github.com/ros-infrastructure/rosdep/issues/752
workaround for rosdep bug at https://github.com/osrf/docker_images/pull/398

PR #189, opened 2020-04-04
Bump checkout to v2

mikaelarguedas:
Checkout v1 has multiple issues, some that are causing build failures:
- unable to checkout on job rerun https://github.com/actions/checkout/issues/23
- not filling PR number correctly on job rerun https://github.com/actions/checkout/issues/58#issuecomment-589447479

This should hopefully address those.

Second commit is unrelated and for nicer prints in the job console output

@kyrofa FYI

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,14 +45,18 @@ jobs:
     container: osrf/ros2:nightly
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
merging this should also reconcile the state of coverage on master and allow to have proper coverage diff in subsequent PRs

PR #191, opened 2020-04-07
api: reorganize permission API

kyrofa:
This PR continues the reorganization of #188 by extracting the permission API into its own module.

### DIFF
--- a/sros2/sros2/api/_permission.py
+++ b/sros2/sros2/api/_permission.py
@@ -1,0 +1,73 @@
+# Copyright 2016-2019 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


sros2/sros2/api/_permission.py
```diff
@@ -0,0 +1,73 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+
+from lxml import etree
+
+import rclpy
```
mikaelarguedas, line 19: As we don't use anything else from rclpy, could we keep `from rclpy.utilities import get_rmw_implementation_identifier` instead?

kyrofa: Of course, sorry, done.

PR #192, opened 2020-04-07
api: reorganize key API

kyrofa:
This PR continues the reorganization of #188 by extracting the key API into its own module.

### DIFF
--- a/sros2/sros2/api/_key.py
+++ b/sros2/sros2/api/_key.py
@@ -1,0 +1,139 @@
+# Copyright 2019-2020 Canonical Ltd
+# Copyright 2016-2019 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


sros2/sros2/api/_key.py
```diff
@@ -0,0 +1,139 @@
+# Copyright 2019-2020 Canonical Ltd
```
ruffsl, line 1: Nit: I think @mikaelarguedas commented on this same copyright notice from last PR.

kyrofa: This one should be legit; I wrote some of these functions and modified others.

sros2/sros2/api/_key.py
```diff
@@ -0,0 +1,139 @@
...
+        raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), keystore_path)
+    if not os.path.isdir(contexts_path):
+        return True
+    for name in os.listdir(contexts_path):
+        if os.path.isdir(os.path.join(contexts_path, name)):
+            print(name)
+    return True
+
+
+def _is_key_name_valid(name):
```
artivis, line 113: Given that `create_key` is public, maybe this is worth being public too?

mikaelarguedas: My understanding from https://github.com/ros2/sros2/pull/188#issuecomment-609054839 is that all API not currently used outside the module is made non-public.
After all the refactor PRs are merged we'll do another pass to decide which additional ones to make public

artivis: Alright sounds good.

PR #195, opened 2020-04-08
api: reorganize artifact generation API

kyrofa:
This PR continues the reorganization of #188 by extracting the artifact generation API into its own module.

### DIFF
--- a/sros2/sros2/verb/generate_artifacts.py
+++ b/sros2/sros2/verb/generate_artifacts.py
@@ -23,7 +23,7 @@ except ImportError:
     def FilesCompleter(*, allowednames, directories):
         return None
 
-from sros2.api import generate_artifacts
+from sros2.api import _artifact_generation
 from sros2.verb import VerbExtension
 
 
@@ -44,7 +44,7 @@ class GenerateArtifactsVerb(VerbExtension):
 
     def main(self, *, args):
         try:
-            success = generate_artifacts(
+            success = _artifact_generation.generate_artifacts(
                 args.keystore_root_path, args.security_contexts, args.policy_files)
         except FileNotFoundError as e:
             raise RuntimeError(str(e))
--- a/sros2/sros2/api/__init__.py
+++ b/sros2/sros2/api/__init__.py
@@ -13,12 +13,6 @@
 # limitations under the License.
 
 from collections import namedtuple
-import os
-import sys
-
-from sros2.policy import load_policy
-
-from . import _key, _keystore, _permission, _policy
 
 HIDDEN_NODE_PREFIX = '_'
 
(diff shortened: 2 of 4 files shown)


sros2/sros2/verb/generate_artifacts.py
```diff
@@ -44,8 +48,43 @@ def add_arguments(self, parser, cli_name):
...
         try:
-            success = generate_artifacts(
+            success = _generate_artifacts(
                 args.keystore_root_path, args.security_contexts, args.policy_files)
         except FileNotFoundError as e:
             raise RuntimeError(str(e))
         return 0 if success else 1
+
+
+def _generate_artifacts(keystore_path=None, identity_names=[], policy_files=[]):
```
kyrofa, line 58: Note that using lists as default values should probably be fixed in a follow-up.

kyrofa:
@mikaelarguedas you'll notice I'm not using any git tricks on this one, because I don't know any that work when moving code into a file that already exists. Do you?

sros2/sros2/verb/generate_artifacts.py
```diff
@@ -44,8 +48,43 @@ def add_arguments(self, parser, cli_name):
...
+            if identity_name not in identity_names:
+                if not _key.create_key(keystore_path, identity_name):
+                    return False
+            policy_element = _policy.get_policy_from_tree(identity_name, policy_tree)
+            _permission.create_permissions_from_policy_element(
+                keystore_path, identity_name, policy_element)
+    return True
+
+
+def _get_keystore_path_from_env():
```
mikaelarguedas, line 85: this would be more suited in  `_utilities` and called from this file. This could come in handy for other verbs (e.g. if we make them use kwargs as suggested by @ruffsl)

sros2/sros2/verb/generate_artifacts.py
```diff
@@ -44,8 +48,43 @@ def add_arguments(self, parser, cli_name):
 
     def main(self, *, args):
         try:
-            success = generate_artifacts(
+            success = _generate_artifacts(
```
mikaelarguedas, line 51: I think of the content of the "verb" files as not public API. Like it's used by that verb but it it not in the API module.
It would be convenient for function like generate_artifacts to be in the api module to be usable by code outside sros2.
Right now multiple pieces of code (`sros2_cmake` and `test_security`) call the CLI, but if any was in python they could use this API instead

kyrofa: Alright I'll rework this to be more similar to the others.

mikaelarguedas:
> you'll notice I'm not using any git tricks on this one, because I don't know any that work when moving code into a file that already exists. Do you?

That's where my git-fu stops as well. Git is alledgedly able to track such copies using `git blame -C` but we won't be able to track this through the github blame interface. In that case we'll just have to live with following the change by hand. It's still moved in the same commit so not that hard to track

sros2/sros2/verb/generate_artifacts.py
kyrofa: Alright, done. Refactored ever so slightly to pull the variable out into a global like the others in utility.

sros2/sros2/verb/generate_artifacts.py
kyrofa: Done.

kyrofa:
Alright this should be ready for another pass.

PR #196, opened 2020-04-08
api: reorganize policy generation API

kyrofa:
Extract policy generation API from api's `__init__.py` and move into verb. Nothing else uses it, and the verb includes a lot of related API on its own.

### DIFF
--- a/sros2/sros2/verb/generate_policy.py
+++ b/sros2/sros2/verb/generate_policy.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from collections import namedtuple
 import os
 import sys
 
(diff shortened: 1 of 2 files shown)


sros2/sros2/verb/generate_policy.py
```diff
@@ -125,23 +124,63 @@ def main(self, *, args):
...
                 if request_services:
                     self.add_permission(
                         profile, 'service', 'request', 'ALLOW', request_services, node_name)
 
         with open(args.POLICY_FILE_PATH, 'w') as stream:
             dump_policy(policy, stream)
         return 0
+
+
+def _get_node_names(*, node, include_hidden_nodes=False):
```
kyrofa, line 149: I don't think these are more generally useful. They're very specific to this verb. Other packages will consume node names/namespaces on their own.

mikaelarguedas: Agreed

PR #201, opened 2020-04-23
update basic tutorials for foxy

mikaelarguedas:

### DIFF
--- a/SROS2_MacOS.md
+++ b/SROS2_MacOS.md
@@ -78,7 +78,6 @@ ros2 security create_keystore demo_keys
 #### Generate keys and certificates for the talker and listener nodes
 
 ```bash
-
 ros2 security create_key demo_keys /talker_listener/talker
 ros2 security create_key demo_keys /talker_listener/listener
 ```
--- a/SROS2_MacOS.md
+++ b/SROS2_MacOS.md
@@ -78,8 +78,9 @@ ros2 security create_keystore demo_keys
 #### Generate keys and certificates for the talker and listener nodes
 
 ```bash
-ros2 security create_key demo_keys /talker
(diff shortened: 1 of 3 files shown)


SROS2_MacOS.md
```diff
@@ -78,8 +78,9 @@ ros2 security create_keystore demo_keys
 #### Generate keys and certificates for the talker and listener nodes
 
 ```bash
-ros2 security create_key demo_keys /talker
-ros2 security create_key demo_keys /listener
+
```
ruffsl, line 81: nit: no white space needed.

mikaelarguedas: :+1: see https://github.com/ros2/sros2/pull/201/commits/2d2ab6deec51dcb6a55d1ce6d9507879aea4b4ce

mikaelarguedas:
> There are some parts of the tutorial that still sound like each Node can have independent security settings, e.g.:

@ivanpauno I'm not sure what the best wording for this would be, we could replace "node" with "process" but apart from that I'm not sure how to add correctness without adding complexity.
Did you have a wording suggestion in mind ?

ivanpauno:
> @ivanpauno I'm not sure what the best wording for this would be, we could replace "node" with "process" but apart from that I'm not sure how to add correctness without adding complexity.
Did you have a wording suggestion in mind ?

Not much more than that, my idea was to replace `node` with `process` everywhere (or almost everywhere).

I think it can be done in a follow-up if you want, as this makes the tutorial work again.

mikaelarguedas:
I gave that an unsatisfactory try so I'll just merge this as is as suggested to make fix the tutorial and revisit the wording at a later date. thanks @ivanpauno for the feedback

PR #214, opened 2020-05-14
Make use of ros_testing to test policy generation.

hidmic:
This pull request combines #168 and #212 to increase test coverage and reduces test flakiness for sros2 `generate_policy` verb.

### DIFF
--- a/sros2/package.xml
+++ b/sros2/package.xml
@@ -22,6 +22,8 @@
   <test_depend>ament_mypy</test_depend>
   <test_depend>ament_pep257</test_depend>
   <test_depend>python3-pytest</test_depend>
+  <!-- Single entrypoint to depend on launch based testing -->
+  <test_depend>ros_testing</test_depend>
   <test_depend>test_msgs</test_depend>
 
   <export>
--- a/sros2/test/sros2/commands/security/verbs/fixtures/client_service_node.py
+++ b/sros2/test/sros2/commands/security/verbs/fixtures/client_service_node.py
@@ -1,0 +1,51 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import rclpy
+from rclpy.node import Node
+
+import test_msgs.srv
+
+
+class ClientServiceNode(Node):
+
+    def __init__(self, name='client_srv_node'):
+        super().__init__(name)
+        self.client = self.create_client(test_msgs.srv.Empty, '~/client')
+        self.service = self.create_service(
+            test_msgs.srv.Empty, '~/server', lambda request, response: response
+        )
+
+    def destroy_node(self):
+        self.client.destroy()
+        self.service.destroy()
+        super().destroy_node()
+
+
+def main(args=None):
+    rclpy.init(args=args)
+
+    node = ClientServiceNode()
+
+    try:
+        rclpy.spin(node)
+    except KeyboardInterrupt:
+        print('node stopped cleanly')
+    finally:
+        node.destroy_node()
+        rclpy.shutdown()
+
+
+if __name__ == '__main__':
+    main()
--- a/sros2/test/sros2/commands/security/verbs/test_generate_policy.py
+++ b/sros2/test/sros2/commands/security/verbs/test_generate_policy.py
@@ -17,6 +17,7 @@ import itertools
 import os
 import sys
 import tempfile
+import unittest
 
 from launch_ros.actions import Node
 
@@ -37,6 +38,7 @@ from utilities.sros2_cli_test_case import SROS2CLITestCase  # noqa: E402
 
 
 @pytest.mark.rostest
+@unittest.skipIf(os.name == 'nt', 'Flaky on Windows')
 @launch_testing.parametrize('rmw_implementation,use_daemon', itertools.product(
     get_available_rmw_implementations(), (True, False)
 ))
--- a/sros2/test/sros2/commands/security/verbs/test_generate_policy.py
+++ b/sros2/test/sros2/commands/security/verbs/test_generate_policy.py
@@ -1,4 +1,5 @@
 # Copyright 2019 Canonical Ltd
+# Copyright 2020 Open Source Robotics Foundation, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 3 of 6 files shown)


hidmic:
Both `test_nightly` CI jobs are failing because `librmw_connext_cpp.so` cannot be found. I can only guess why the focal PR job failed, it seems like discovery latency to me. Bumped spin times from 1.0 to 2.0 seconds.

mikaelarguedas:
> Both test_nightly CI jobs are failing because librmw_connext_cpp.so cannot be found. 

Noticed this too and confirmed that using  an image with connext installed will fix the failing test: https://github.com/ros2/sros2/runs/677227888.

sros2/test/sros2/commands/security/verbs/test_generate_policy.py
```diff
@@ -12,133 +13,155 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import itertools
 import os
+import sys
 import tempfile
 
+from launch_ros.actions import Node
```
mikaelarguedas, line 21: this should be added as a dependency in the package.xml

sros2/test/sros2/commands/security/verbs/test_generate_policy.py
```diff
@@ -12,133 +13,155 @@
...
-
-def test_generate_policy_no_policy_file(capsys):
-    with pytest.raises(SystemExit) as e:
-        cli.main(argv=['security', 'generate_policy'])
-        assert e.value.code != 0
-    stderr = capsys.readouterr().err.strip()
-    assert 'following arguments are required: POLICY_FILE_PATH' in stderr
+
+from utilities.sros2_cli_test_case import generate_sros2_cli_test_description
+from utilities.sros2_cli_test_case import SROS2CLITestCase
```
mikaelarguedas, line 35: Should these "utilities" imports be prefixed with a '.' or using a full name ?

sros2/test/sros2/commands/security/verbs/test_generate_policy_no_nodes.py
```diff
@@ -0,0 +1,63 @@
+# Copyright 2019 Canonical Ltd
+# Copyright 2020 Open Source Robotics Foundation, Inc.
```
mikaelarguedas, line 2: File has multiple copyright statements but no history

sros2/test/sros2/commands/security/verbs/utilities/sros2_cli_test_case.py
```diff
@@ -0,0 +1,94 @@
...
+
+from launch import LaunchDescription
+from launch.actions import ExecuteProcess
+from launch.actions import TimerAction
+
+import launch_testing
+import launch_testing.asserts
+import launch_testing.markers
+import launch_testing.tools
+import launch_testing_ros.tools
```
mikaelarguedas, line 26: these dependencies should be listed in the package.xml

sros2/package.xml
```diff
@@ -22,6 +22,7 @@
   <test_depend>ament_mypy</test_depend>
   <test_depend>ament_pep257</test_depend>
   <test_depend>python3-pytest</test_depend>
+  <test_depend>ros_testing</test_depend>
```
mikaelarguedas, line 25: I don't see the "ros_testing" string anywhere in this PR. Is this dependency being used ?

sros2/test/sros2/commands/security/verbs/test_generate_policy.py
```diff
@@ -12,133 +13,155 @@
...
 # limitations under the License.
 
+import itertools
 import os
+import sys
 import tempfile
 
+from launch_ros.actions import Node
+
+import launch_testing
```
mikaelarguedas, line 23: same here

sros2/test/sros2/commands/security/verbs/fixtures/client_service_node.py
```diff
@@ -0,0 +1,51 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import rclpy
+from rclpy.node import Node
+
+import test_msgs.srv
+
+
+class ClientServiceNode(Node):
```
kyrofa, line 21: Can these be turned into an actual pytest fixture on a conftest.py somewhere?

sros2/test/sros2/commands/security/verbs/utilities/sros2_cli_test_case.py
```diff
@@ -0,0 +1,94 @@
...
+        ExecuteProcess(
+            cmd=['ros2', 'daemon', 'stop'],
+            name='daemon-stop',
+            on_exit=fixture_actions,
+            additional_env=additional_env
+        ),
+    ])
+
+
+class SROS2CLITestCase(unittest.TestCase):
```
kyrofa, line 59: The existing tests in this project are all pytest. Why are we using unittest here?

sros2/test/sros2/commands/security/verbs/test_generate_policy.py
```diff
@@ -12,133 +13,155 @@
...
+            ReadyToTest()
+        ],
+        rmw_implementation=rmw_implementation,
+        use_daemon=use_daemon
+    ), locals()
+
+
+class TestSROS2GeneratePolicyVerb(SROS2CLITestCase):
+
+    @launch_testing.markers.retry_on_failure(times=3)
```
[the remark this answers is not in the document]
hidmic: > Does this mean the test will be retried automatically up to 3 times and not report the failure if it passes at least one of these times?

Correct.

> How is this controlled externally ? Is it possible to not rerun these by default ?

There's no external control. 

We can definitely remove those decorators and try. Truth is that both chosen delay and maximum retries here are completely arbitrary. There's no statistical survey backing up these upper bounds.

sros2/test/sros2/commands/security/verbs/test_generate_policy.py
hidmic: `ros_testing` will bring the entire `launch(_ros)?` and `launch_testing(_ros)?` ecosystem (that's the point of it), but we can be explicit if you'd like.

sros2/test/sros2/commands/security/verbs/test_generate_policy_no_nodes.py
hidmic: That's a very good point. It is work derived from a source file that was originally developed by Canonical. Do tell how we should proceed with this.

sros2/test/sros2/commands/security/verbs/utilities/sros2_cli_test_case.py
hidmic: Same as above.

sros2/package.xml
hidmic: Same as above.

sros2/test/sros2/commands/security/verbs/test_generate_policy.py
hidmic: Same as above.

sros2/test/sros2/commands/security/verbs/fixtures/client_service_node.py
hidmic: Short answer is no, because `test_generate_policy*.py` is no longer a `pytest`, but a `pytest`-run `launch` test.

sros2/test/sros2/commands/security/verbs/utilities/sros2_cli_test_case.py
hidmic: Because `test_generate_policy*.py` is no longer a `pytest`, but a `pytest`-run `launch` test. And `launch_testing` is built on top of `unittest`. There's a desire to achieve a stronger integration between `launch_testing` and `pytest` (even turning the former into a plugin for the latter, see https://github.com/ros2/launch/issues/405), but as of today, we can only include `launch` tests in `pytest` runs.

sros2/test/sros2/commands/security/verbs/test_generate_policy.py
hidmic: A '.' prefix suggests that the test lives in a Python module/package which is not the case. I'm not sure what you mean by full name. 

I can (and probably should) ensure the sibling `utilities` module is used by tweaking `sys.path`.

hidmic: See https://github.com/ros2/sros2/pull/214/commits/30b0fe341e643fa671b45a79cd6ab4dea4a24d31.

sros2/test/sros2/commands/security/verbs/test_generate_policy.py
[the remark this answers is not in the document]
hidmic: See https://github.com/ros2/sros2/pull/214/commits/30b0fe341e643fa671b45a79cd6ab4dea4a24d31.

sros2/test/sros2/commands/security/verbs/test_generate_policy_no_nodes.py
kyrofa: Should it actually share git history with another file? Or is it more of a "I copied that file and turned it into something else" situation?

sros2/test/sros2/commands/security/verbs/utilities/sros2_cli_test_case.py
kyrofa: Alright, thank you for explaining. I really dislike mixing the two, here. It's complicated enough having both of them in existence and sharing space in my head-- having them both used in the same project will cause no end of confusion. Is there any other way to do this?

sros2/test/sros2/commands/security/verbs/test_generate_policy.py
mikaelarguedas: In general we aim for "declare everything you use directly" to avoid the effects of dependencies disappearing under the hood and breaking us. But if the raison-d'etre of the `ros_testing` package is to give a single entrypoint for all things testing and always provide all these dependencies then I guess that's fine.

Can you please add a comment to that effect in the package.xml ? this will avoid future us from pruning the dependency

sros2/test/sros2/commands/security/verbs/test_generate_policy.py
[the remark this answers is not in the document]
mikaelarguedas: Great thanks, if it happens to still be flakey after these imrpovements feel free to open an issue and we'll reconsider the retries / tweaking timeouts

sros2/test/sros2/commands/security/verbs/test_generate_policy.py
mikaelarguedas: :+1:

sros2/test/sros2/commands/security/verbs/test_generate_policy_no_nodes.py
hidmic: It's a modified copy of `test_generate_policy.py`.

sros2/test/sros2/commands/security/verbs/utilities/sros2_cli_test_case.py
hidmic: There's not, assuming that testing against multiple RMW implementations is a requirement (I understand it is based on https://github.com/ros2/sros2/pull/212#issuecomment-628245056) and that we're not looking forward to recreate `launch` and `launch_testing` using `subprocess` and `pytest` just for this one test. I'd rather not do that.

hidmic:
CI up to `sros2` and `test_security`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10824/)

hidmic:
I'm not happy with https://github.com/ros2/sros2/pull/214/commits/3e1bc3951050f1ef3d5844d80030dd4bd3c136e9, for a few different reasons. But I'm unable to reproduce Windows CI failures in a Windows VM. CC @jacobperron

hidmic:
Alright, I corrected the commit history to pass DCO and make sense. CI's green, though tests are being skipped on Windows. FYI @mikaelarguedas @kyrofa. @jacobperron how shall we proceed?

mikaelarguedas:
Regarding the move away from pytest and mixing unittest and pytest in the same package, I'm not a fan but it wouldn't be a blocker for me.

However, the fact that this seems to have pretty consistently failed CI on this repo makes me doubt this is having the intended effect of reducing flakeyness. So IMO it can't be merged as is.

I was hoping that the flakeyness was due to adding connext into the mix but several failing tests seem to be related to fastrtps(static or dynamic) so it connext doesn't seem to be the culprit.

We could scale back this to just provide a single RMW with daemon and "long" discovery time. if that proves to reduce flakeyness I'm in favor of merging that as it'll keep covering the cases users will face when using the CLI while improving CI stability.

hidmic:
I have to circle back and see what's happening with Fast-RTPS now. In all cases, however, the issue (it seems) remains to be discovery latency. Testing with and without a daemon and against multiple RMW implementations only made it more visible. 

I agree that to merge as-is isn't viable. I'm not sure I'd test less either just to get jobs to pass. Maybe we need to start ensuring preconditions (in this case, ROS graph state) hold instead of hoping they will within an arbitrary timeout.

hidmic:
Had to bump up the timeout for `ros2 security generate_policy` execution. I've seen it taking ~13 seconds on a pretty much idle Windows Server VM.

hidmic:
Ok, https://github.com/ros2/sros2/pull/214/commits/b2a8bcbf7da23ac23bd5902c56b84b4fdae850ce is almost insane. Those that timeout are the ones that use the daemon, and so they have the added overhead of XMLRPC calls. Still, it's ridiculous.

hidmic:
CI up to `sros2` and `test_security`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11116/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6420/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9070/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11042/)

@mikaelarguedas @kyrofa PTAL, I think I've managed to remove most flakiness sources. I'm a bit baffled though by the time it takes the verb to complete in Windows.

mikaelarguedas:
> Ok, b2a8bcb is almost insane. Those that timeout are the ones that use the daemon, and so they have the added overhead of XMLRPC calls. Still, it's ridiculous.

This seems indeed incredible that it takes more than 20seconds to get the accurate state of a graph with 1 node and one Publisher/Subscriber pair... Is the same happening for other tests doing similar things ? or is there something wrong with the way testing is done here ?

> I think I've managed to remove most flakiness sources. 

As far as CI goes on this repo the latest iterations of this PR seem to yield only green jobs which is a big improvement compared to last review :+1:

hidmic:
> Is the same happening for other tests doing similar things ? or is there something wrong with the way testing is done here ?

I think almost all CLI tests are disabled on Windows due to https://github.com/ros2/build_farmer/issues/248.

At any rate, as I mentioned in https://github.com/ros2/sros2/pull/214#issuecomment-642903757, I do see `ros2 security generate_policy` taking 10-15 seconds on a Windows VM.

mikaelarguedas:
> I do see ros2 security generate_policy taking 10-15 seconds on a Windows VM by calling it manually

I missed that bit. This seems very long too.. On linux its more on the second scale.

I'm +1 for this change as it increases use cases covered, does not degrade CI here and (IIUC) decreases CI flakyness on ci.ros2.org.

I'll leave it up to @kyrofa for the second review

hidmic:
@kyrofa friendly ping.

PR #215, opened 2020-05-15
use nightly image including all 3 RMWs

mikaelarguedas:
This will unlock PRs adding tests for multiple RMW implementations such as https://github.com/ros2/sros2/pull/214

For now no test expect security DDS plugins to be installed. If that becomes the case we would need to either blacklist connext or find a way for the tests to fetch a license with security support and install the security plugins.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,6 +9,8 @@ on:
 jobs:
   test_latest:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
> "this doesn't include connext security plugins" as a comment in the YAML so we don't lose that tidbit?

Sure thing, added it in https://github.com/ros2/sros2/pull/215/commits/f752a734335b0fea5f180199a014b3973ea9df22 for both jobs

kyrofa:
Thank you!

Issue #183, opened 2020-03-26
[Dashing/Eloquent] List keys doesn't list nested keys


ivanpauno:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04 (should happen in any)
- Installation type:
  - from source (should happen in any)
- Version or commit hash:
  - master
- DDS implementation:
  - N/A
- Client library (if applicable):
  - N/A

#### Steps to reproduce issue

```
ros2 security generate_artifacts -k tmp -n /node1 /ns/node1 /ns/node2
ros2 security list_keys tmp
```
#### Expected behavior
Output:
```
/node1
/ns/node1
/ns/node2
```

#### Actual behavior
Output:
```
node1
ns
```

#### Additional information

This should be fixed for Dashing/Eloquent, and also for Foxy after https://github.com/ros2/sros2/pull/177 is merged.
Add a regression test too.

hidmic:
@ros2/security_working_group are you planning to take a stab at this one in the near term e.g. before Foxy feature freeze?

kyrofa:
Do the still-in-flux design updates impact this at all?

ivanpauno:
> Do the still-in-flux design updates impact this at all?

The fix for `Dashing` and `Eloquent` aren't blocked by those changes.
About `Foxy`, maybe `contexts` folder is renamed to `enclaves`, but it isn't a big deal to update the folder name in a follow up.

ivanpauno:
Reopening to track the fix for Dashing/Eloquent

mikaelarguedas:
Backported and released
Dashing: https://github.com/ros2/sros2/pull/223 https://github.com/ros/rosdistro/pull/25499
Eloquent: https://github.com/ros2/sros2/pull/222 https://github.com/ros/rosdistro/pull/25492

PR #219, opened 2020-05-30
Fix list keys verb

mikaelarguedas:
Fixes https://github.com/ros2/sros2/issues/183
setup: 
`ros2 security generate_artifacts -k keystore -e /foo /foobar/baz /bar /a/b/c/deep`

before:
```
root@5ab52a51093c:/ws# ros2 security list_keys keystore/
foobar
bar
foo
a

```
after:
```
root@5ab52a51093c:/ws# ros2 security list_keys keystore/
/a/b/c/deep
/bar
/foo
/foobar/baz
```

Note: a similar fix could be backported to Eloquent and Dashing

### DIFF
--- a/sros2/sros2/api/_key.py
+++ b/sros2/sros2/api/_key.py
@@ -111,7 +111,7 @@ def list_keys(keystore_path):
     p = pathlib.Path(enclaves_path)
     key_file_paths = sorted(p.glob('**/key.pem'))
     for key_file_path in key_file_paths:
-        print('/{}'.format(key_file_path.parent.relative_to(enclaves_path)))
+        print('/{}'.format(key_file_path.parent.relative_to(enclaves_path).as_posix()))
     return True
 
 
--- a/sros2/sros2/api/_key.py
+++ b/sros2/sros2/api/_key.py
@@ -15,6 +15,7 @@
 
 import errno
 import os
+import pathlib
 
 from cryptography import x509
 from cryptography.hazmat.backends import default_backend as cryptography_backend
@@ -103,12 +104,14 @@ def create_key(keystore_path, identity):
 def list_keys(keystore_path):
     enclaves_path = _keystore.get_keystore_enclaves_dir(keystore_path)
     if not os.path.isdir(keystore_path):
-        raise FileNotFoundError(errno.ENOENT, os.strerror(errno.ENOENT), keystore_path)
+        raise FileNotFoundError(
+            errno.ENOENT, os.strerror(errno.ENOENT), keystore_path)
     if not os.path.isdir(enclaves_path):
         return True
-    for name in os.listdir(enclaves_path):
-        if os.path.isdir(os.path.join(enclaves_path, name)):
-            print(name)
+    p = pathlib.Path(enclaves_path)
+    key_file_paths = sorted(p.glob('**/key.pem'))
+    for key_file_path in key_file_paths:
+        print('/{}'.format(key_file_path.parent.relative_to(enclaves_path)))
     return True
 
 
--- a/sros2/test/sros2/commands/security/verbs/test_list_keys.py
+++ b/sros2/test/sros2/commands/security/verbs/test_list_keys.py
@@ -32,7 +32,7 @@ def test_list_keys(capsys):
 
         # Now verify that the key we just created is included in the list
         assert cli.main(argv=['security', 'list_keys', keystore_dir]) == 0
-        assert (capsys.readouterr().out.strip() == '\n'.join(sorted(enclave_names)))
+        assert capsys.readouterr().out.strip() == '\n'.join(sorted(enclave_names))
 
 
 def test_list_keys_no_keys(capsys):
--- a/sros2/test/sros2/commands/security/verbs/test_list_keys.py
+++ b/sros2/test/sros2/commands/security/verbs/test_list_keys.py
@@ -20,17 +20,19 @@ from sros2.api import _key, _keystore
 
 
 def test_list_keys(capsys):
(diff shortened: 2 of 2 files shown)


mikaelarguedas:
@kyrofa CI on this PR is another occurrence of test_security tests being flaky https://github.com/ros2/sros2/pull/219/checks?check_run_id=722938561

sros2/sros2/api/_key.py
```diff
@@ -103,12 +104,14 @@ def create_key(keystore_path, identity):
...
+            errno.ENOENT, os.strerror(errno.ENOENT), keystore_path)
     if not os.path.isdir(enclaves_path):
         return True
-    for name in os.listdir(enclaves_path):
-        if os.path.isdir(os.path.join(enclaves_path, name)):
-            print(name)
+    p = pathlib.Path(enclaves_path)
+    key_file_paths = sorted(p.glob('**/key.pem'))
+    for key_file_path in key_file_paths:
+        print('/{}'.format(key_file_path.parent.relative_to(enclaves_path)))
```
kyrofa, line 114: Should this use `os.sep`?

ruffsl: If it uses the same / used in designating the enclave name path, then the stdout could be piped for more commands involving the enclave name.

kyrofa: That's a good idea, @ruffsl. I actually think my initial comment is wrong; I believe pathlib will use forward slashes regardless of the OS, but your suggestion requires a consistency in path handling that I don't think we have right now. We should take a pass on converting things to pathlib and then we can use forward slashes reliably. That's outside this PR, of course.

Arnatious: Running the code as-is on windows, with a file `.\foo\bar\key.pem', gave a printout `/foo\bar`

Arnatious: Pathlib can read './foo\\bar` just fine and resolve it to the same path

kyrofa: Thanks @Arnatious. I don't think we want to print `/foo\bar` though. I'm back to my original suggestion of using `os.sep`, or we can use `as_posix()` when printing them to stdout.

mikaelarguedas: Good point the current bewhavior is not what we'd want.

> I'm back to my original suggestion of using os.sep

It depends what we think the function returns.
If it is to give print a list of enclave names, the fully qualified name should be used (forward slashes regardless of the platform)
If it is to print a list of enclave paths it should us the one from the OS (and likely not use an absolute path but a relative one ?)

I'm personally leaning towards the former (return the fully qualified name that's used in the certs/permission files)

kyrofa: > I'm personally leaning towards the former (return the fully qualified name that's used in the certs/permission files)

Agreed. `as_posix()`, then?

mikaelarguedas: > Agreed. as_posix(), then?

:+1: for me, I'd like to hear how other feel about it.

---

One could argue that if we are to return the enclave name it should be extracted from the files (e.g. from permissions.xml) rather than inferred from location

kyrofa: > One could argue that if we are to return the enclave name it should be extracted from the files (e.g. from permissions.xml) rather than inferred from location

Then that person should also be saying the lookup strategy in rcl is wrong :stuck_out_tongue: .

mikaelarguedas: :smile: for sure, I'll be honest and admit I'm not sure how rcl deals with it these days. I believe it takes the path with forward slashes and automagically resolves it the right way on windows.

It just seems less hacky to do "read value from file all files matching pattern" than "let's find the relative path of all the files matching that name but export it as a posix path and pretend it's absolute but as if the enclaves folder was the root".

On the other hand it's much easier to update the patch to use `as_posix()` and it fixes the current issue so I'm happy to do that as an incremental improvement :stuck_out_tongue:

kyrofa: Haha, yeah I'm not sure about that part either, but it' still path-based, regardless of the slashes, so we might as well do the same thing here. Otherwise it could be pretty confusing to print the contents of the files but then have RCL not actually match those up on the backend, you know?

mikaelarguedas: fair enough :+1:

ruffsl: I'm not sure where you're going, but I always saw the list command as a numerating a list of fully qualified enclave names. If the certificates were generated by SROS, then we could also look at the certificate common name, but those aren't guaranteed to match it's location as it can be moved in the key store. But I think we should stick with extrapolating the enclave name from its placement in the key store, as that's how RCL will resolve the enclave name within file system type key stores.

kyrofa: > But I think we should stick with extrapolating the enclave name from its placement in the key store, as that's how RCL will resolve the enclave name within file system type key stores.

Sorry if I was unclear, that ^ is what I meant.

mikaelarguedas: added `as_posix` in https://github.com/ros2/sros2/pull/219/commits/23e41bf00a85fb7055c62b4428e6d81c945e3cec

mikaelarguedas:
With some non-Linux CI to see the effects:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11003/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6339/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8963/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10929/)

PR #220, opened 2020-05-30
[tutorial] fix copy-n-paste error on enclave path

mikaelarguedas:
Also installs subversion on Linux for the part of the tutorial that needs it.


mikaelarguedas:
test failures unrelated (tracked at https://github.com/ros2/sros2/issues/221)

Issue #221, opened 2020-05-30
[test_security] Secure communication tests are unstable causing CI to fail sometimes


mikaelarguedas:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu Focal
- Installation type:
  - source
- Version or commit hash:
  - master
- DDS implementation:
  - Fast-RTPS (static and dynamic)
- Client library (if applicable):
  - rclcpp

Since https://github.com/ros2/sros2/pull/208 the end-to-end tests of sros2 are enabled on the CI of this repo (daily master branch test + PR testing).

Some tests time out making CI fail, this seems unrelated to the changes being tested, some examples of this:
- Daily build: https://github.com/ros2/sros2/runs/710709221?check_suite_focus=true
- PR builds:
  - https://github.com/ros2/sros2/runs/722938561
  - https://github.com/ros2/sros2/pull/220/checks?check_run_id=723326846
  - https://github.com/ros2/sros2/pull/220/checks?check_run_id=723328796

While the code is located at https://github.com/ros2/system_tests/tree/master/test_security I'm reporting it here because that's the repo watched by he sros2 maintainer and the test failures are happening on the CI of this repo.

Looking at the ROS2 CI from ci.ros2.org over the last week it seems to fail very often but only on linux and only for fastrtps. It goes unnoticed because tests are ran multiple times:
## Linux:
- [May 30, 2020](https://ci.ros2.org/view/nightly/job/nightly_linux_release/1558)
```
  33 - test_secure_publisher_subscriber__Empty__rmw_fastrtps_cpp__secure_comm_3 (Timeout)
  43 - test_secure_publisher_subscriber__UnboundedSequences__rmw_fastrtps_cpp__secure_comm_2 (Timeout)
```
- [May 29, 2020](https://ci.ros2.org/view/nightly/job/nightly_linux_release/1557)
```
  41 - test_secure_publisher_subscriber__UnboundedSequences__rmw_fastrtps_cpp__secure_comm_0 (Timeout)
```
- [May 28, 2020](https://ci.ros2.org/view/nightly/job/nightly_linux_release/1556)
```
  43 - test_secure_publisher_subscriber__UnboundedSequences__rmw_fastrtps_cpp__secure_comm_2 (Timeout)
```
- [May 27, 2020](https://ci.ros2.org/view/nightly/job/nightly_linux_release/1555)
  :heavy_check_mark:
[...]

PR #225, opened 2020-06-22
Remove the use of pkg_resources.

clalancette:
Importing pkg_resources is slow; really slow.  pkg_resources
is also somewhat deprecated.  For both of these reaons, we
would prefer to move away from pkg_resources and start using
importlib instead.

sros2 was using 'resource_filename' from pkg_resources.  The
equivalent in importlib is importlib.resources, but there is a catch.
As of Python 3.8, importlib.resources does not support getting
data from subdirectories.  The workaround is to make each
level of the directory structure a python module, so that the
module name can be looked up.  Python 3.9 will have support for
directories for resources, but that won't help on Ubuntu Focal.
More information is in https://gitlab.com/python-devs/importlib_resources/issues/58

See ros2/ros2#919 for a lot more details on why we want to do this.

### DIFF
--- a/sros2/sros2/policy/__init__.py
+++ b/sros2/sros2/policy/__init__.py
@@ -16,45 +16,47 @@ import os
 
 from lxml import etree
 
-import pkg_resources
+try:
+    import importlib.resources as importlib_resources
+except ModuleNotFoundError:
+    # (clalancette): We have to ignore the type, otherwise mypy throws a warning.
+    # See https://github.com/python/mypy/issues/1153 for details.
+    import importlib_resources  # type: ignore
 
 POLICY_VERSION = '0.2.0'
 
 
 def get_policy_default(name):
-    return pkg_resources.resource_filename(
-        package_or_requirement='sros2',
-        resource_name=os.path.join('policy', 'defaults', name))
+    with importlib_resources.path('sros2.policy.defaults', name) as path:
+        return str(path)
 
 
 def get_policy_schema(name):
-    return pkg_resources.resource_filename(
-        package_or_requirement='sros2',
-        resource_name=os.path.join('policy', 'schemas', name))
+    with importlib_resources.path('sros2.policy.schemas', name) as path:
+        return str(path)
 
 
 def get_policy_template(name):
-    return pkg_resources.resource_filename(
-        package_or_requirement='sros2',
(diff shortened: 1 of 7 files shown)


sros2/sros2/policy/__init__.py
```diff
@@ -16,45 +16,47 @@
 
 from lxml import etree
 
-import pkg_resources
+try:
+    import importlib.resources as importlib_resources
+except ModuleNotFoundError:
+    # (clalancette): We have to ignore the type, otherwise mypy throws a warning.
```
mikaelarguedas, line 22:
```suggestion
    # This fallback is necessary when importlib-resources is installed from pip
    # (clalancette): We have to ignore the type, otherwise mypy throws a warning.
```

clalancette: Makes sense; see b3a53fd.

clalancette:
> Could you please add a comment or TODO to the added `__init__.py` files to explain why they're needed and in what conditions they can be safely removed?

Good call.  I added the comment to all of the __init__ files in b3a53fd.

sros2/sros2/policy/__init__.py
```diff
@@ -16,45 +16,48 @@
...
-        resource_name=os.path.join('policy', 'templates', name))
+    with importlib_resources.path('sros2.policy.templates', name) as path:
+        return str(path)
 
 
 def get_transport_default(transport, name):
-    return pkg_resources.resource_filename(
-        package_or_requirement='sros2',
-        resource_name=os.path.join('policy', 'defaults', transport, name))
+    module = 'sros2.policy.defaults.' + transport
```
mikaelarguedas, line 46: Nit: we're moving towards using f-strings for all string formatting and concatenation in sros2, not a requirement for merging though, (same below)

mikaelarguedas:
CI failure unrelated and due to https://github.com/ros2/sros2/issues/221.

@clalancette feel free to merge this along the other PRs

clalancette:
> CI failure unrelated and due to #221.
> 
> @clalancette feel free to merge this along the other PRs

Sounds good, thanks!

clalancette:
All right, CI changes are merged as part of https://github.com/ros2/ci/pull/482 .  CI is green in https://github.com/ros2/ros2/issues/919 , and I have 3 approvals.  Thanks for the reviews, merging.

PR #230, opened 2020-07-30
Add scope parameter to conftest.py

j-rivero:
Following upstream instructions https://docs.pytest.org/en/stable/fixture.html#scope-sharing-fixtures-across-classes-modules-packages-or-session . Fix #229 

**Testing:**
 * Linux [Build Status](https://ci.ros2.org/job/ci_linux/11704/)
 * Aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/6801/)
 * Win [Build Status](https://ci.ros2.org/job/ci_windows/11654/)
 * OsX [Build Status](https://ci.ros2.org/job/ci_osx/9555/)

PR #232, opened 2020-07-31
Add scope parameter (backport to Foxy)

j-rivero:
Backporting #230 to Foxy to [solve regressions in test](http://build.ros2.org/job/Fci__nightly-release_ubuntu_focal_amd64/166/testReport/junit/sros2.test/test_mypy/test_mypy/).  //cc @jacobperron


mikaelarguedas:
The coverage decrease seems invalid as the coverage of this PR diff is unchanged.
So ignoring the codecov failure and merging as is

PR #231, opened 2020-07-31
add cyclonedds to the list of rmw using graph info topics

mikaelarguedas:
To test if it fixes the issue of CI failing with cyclone and security from https://github.com/ros2/system_tests/pull/408

Necessary since https://github.com/ros2/rmw_cyclonedds/pull/145 particularily this bit https://github.com/ros2/rmw_cyclonedds/blob/3bfe094b60c1a3116b1f2c09c845a745fd240987/rmw_cyclonedds_cpp/src/rmw_node.cpp#L1026-L1035

PR #238, opened 2020-09-02
[foxy-backport] add cyclonedds to the list of rmw using graph info topics (#231)

mikaelarguedas:
As CycloneDDS security has been made available in Foxy (and synced to main already)

### DIFF
--- a/sros2/sros2/api/_permission.py
+++ b/sros2/sros2/api/_permission.py
@@ -23,7 +23,11 @@ from sros2.policy import get_transport_schema, get_transport_template
 from . import _keystore, _policy, _utilities
 
 
-_RMW_WITH_ROS_GRAPH_INFO_TOPIC = ('rmw_fastrtps_cpp', 'rmw_fastrtps_dynamic_cpp')
+_RMW_WITH_ROS_GRAPH_INFO_TOPIC = (
+    'rmw_cyclonedds_cpp',
+    'rmw_fastrtps_cpp',
+    'rmw_fastrtps_dynamic_cpp'
+)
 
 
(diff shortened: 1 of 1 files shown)


kyrofa:
I see that cyclonedds is using [release/0.7.x in foxy](https://github.com/ros2/ros2/blob/foxy/ros2.repos#L29), but if I look at the [changelog](https://github.com/eclipse-cyclonedds/cyclonedds/blob/releases/0.7.x/CHANGELOG.rst) there I don't actually see security-related things. I do on [master](https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/CHANGELOG.rst), but that's not what's in Foxy. I'm obviously missing a piece of the puzzle, can you show me?

mikaelarguedas:
> but if I look at the changelog there I don't actually see security-related things.

I guess the missing piece is that the changelog on the 0.7.X branch is missing the 0.7.0 entry (which is somehow ironic).
The one on master has the changelog of version 0.7.0 [which is the one that's in Foxy](https://github.com/ros/rosdistro/blob/077940b05b7e615436044bc9981e248bd1a0cc1d/foxy/distribution.yaml#L424) and includes the bits about security support.

kyrofa (approved):
Ah I should have looked at rosdistro, okay fair enough, thank you. Great to see this come to Foxy!

mikaelarguedas:
Opened a [PR against the changelogs](https://github.com/eclipse-cyclonedds/cyclonedds/pull/587) to avoid future confusion 

The ROS buildfarm CI failure is due to connectivity issue updating rosdep database and is unrelated to this change

artivis (approved):
lgtm :+1:

PR #240, opened 2020-09-09
Use S/MIME API from cryptography instead of SSL bindings

### LINKED ISSUE #239: Use new S/MIME API in Cryptography
Related to #234, the Cryptography library has removed the SSL bindings for S/MIME signatures in v3.1 that we're using in the SROS2 utilities, but they are working on getting a real, official S/MIME API into a future release. We're stuck on v3.0 until they release the new API, but we need an issue to track this.

kyrofa:
The SROS2 utilities use the Cryptography library, which doesn't officially support S/MIME. As a result, we had to hit the SSL bindings directly to get that functionality. These bindings were removed in v3.1 of Cryptography, which means the SROS2 utilities were stuck on v3.0 at the latest (see https://github.com/pyca/cryptography/issues/5433).

However, Cryptography is working on adding an official S/MIME API which is far easier to use than the raw SSL API. This PR resolves #239 by updating the SROS2 utilities to use it. Note, however, that the S/MIME API has not been merged nor released in Cryptography, so this PR will remain in draft and fail all its tests for a little while.

See also: ros2/ci#507

### DIFF
--- a/sros2/sros2/api/_utilities.py
+++ b/sros2/sros2/api/_utilities.py
@@ -19,9 +19,9 @@ import sys
 
 from cryptography import x509
 from cryptography.hazmat.backends import default_backend as cryptography_backend
-from cryptography.hazmat.bindings.openssl.binding import Binding as SSLBinding
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives import serialization
+from cryptography.hazmat.primitives import smime
 from cryptography.hazmat.primitives.asymmetric import ec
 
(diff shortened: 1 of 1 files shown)


kyrofa:
Once this API is settled, this PR should be refactored to only use it if available, and fallback to the bindings if not. It seems that we will need to support both Cryptography 2.8 and the latest, until rolling has switched to the next Ubuntu LTS beyond 20.04.

clalancette:
This API changed from the time that this PR was opened.  I've now opened up #290 , which supersedes this one.  So closing this out.

PR #243, opened 2020-09-16
no CI against nightly or latest on foxy branch

mikaelarguedas:


kyrofa:
This seems to remove CI completely, can we at least get CI against Foxy?

mikaelarguedas:
The CI from the buildfarm is available for released distros. e.g. on this PR we get the jobs `Fpr__sros2__*` triggered automatically

kyrofa (approved):
Ah, I forgot about tests in the build farm. Works for me, then.

artivis:
So we keep the CI for unreleased branches and hands it off to the build farm for released ones? How does that work wrt Rolling?

artivis (approved):

kyrofa:
> How does that work wrt Rolling?

Good question, I wonder if the build farm supports running tests against rolling now. I suspect it does. I'm also not sure what it really gets us beyond our current tests in master, though.

clalancette:
> Good question, I wonder if the build farm supports running tests against rolling now. I suspect it does. I'm also not sure what it really gets us beyond our current tests in master, though.

It depends on which buildfarm you are talking about.  https://ci.ros2.org/ runs all of the tests from all core packages on all tier 1 platforms nightly.

http://build.ros2.org has PR jobs, which are the ones that show up in the Github "checks" here.  Which PR job is used depends on which branch a PR targets.  For ones that target `foxy`, for instance, the PR job is http://build.ros2.org/view/Fpr/job/Fpr__sros2__ubuntu_focal_amd64/.  For ones that target master, the PR job is: http://build.ros2.org/view/Rpr/job/Rpr__sros2__ubuntu_focal_amd64/ .  In either case, those PR jobs run all of the tests for this particular repository on every PR that is opened, though only on Linux/amd64.

Hope that answers your question.

mikaelarguedas:
> So we keep the CI for unreleased branches and hands it off to the build farm for released ones? How does that work wrt Rolling?

Yes I think that's what we'll do for now. The limiting thing is that (for now both on CI on this repo and for buildfarm PR jobs) we test only ubuntu so it's not covering any of the other platforms..

For Rolling, the release strategy is unclear, for now the version in rolling was automatically released by OR so it is using the branch and package version that was used by the foxy release at the time. We'll need to decide what we release in rolling, if it just tags off master or if we make a rolling branch, on what schedule to release and how we test it (noting that our CI options have the same limitations ATM so onlu ubuntu testing).
Right now I'm leaning towards no dedicated branch, releasing off the default branch and doing thorough testing only when we need to use whats in rolling to kick-off a new distro. So Rolling CI will just be our usual master-branch testing

kyrofa:
> Right now I'm leaning towards no dedicated branch, releasing off the default branch and doing thorough testing only when we need to use whats in rolling to kick-off a new distro. So Rolling CI will just be our usual master-branch testing

I agree with that.

PR #246, opened 2020-12-02
Skip the mypy testing on CentOS 7.

clalancette:
The inline comment has a lot more details on why we need to
do this.

This should fix up the last of the lingering errors on the CentOS job: https://ci.ros2.org/view/nightly/job/nightly_linux-centos_debug/611/

I'll run CI on this once I get agreement that this is the right way to go.

### DIFF
--- a/sros2/test/test_mypy.py
+++ b/sros2/test/test_mypy.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
+
 from ament_mypy.main import main
 import pytest
 
@@ -19,4 +21,23 @@ import pytest
 @pytest.mark.mypy
 @pytest.mark.linter
 def test_mypy():
+    distroname = ''
+    if os.path.exists('/etc/os-release'):
+        with open('/etc/os-release', 'r') as infp:
+            for line in infp:
+                if line.startswith('ID='):
+                    split = line.strip().split('=')
+                    if len(split) == 2:
+                        distroname = split[1].lstrip('"').rstrip('"')
+                    break
+
+    if distroname == 'centos':
+        # CentOS 7 uses a pip installation of importlib_resources to get access
+        # to the importlib_resources API.  Due to a bug
+        # (https://github.com/python/mypy/issues/1153), mypy cannot determine
+        # the API in that case and so throws a warning.  If we see we are on
+        # CentOS, skip the mypy checking; on all other platforms, run the
+        # test as usual.
+        pytest.skip('CentOS 7 does not support mypy checking of importlib properly')
+
     assert main(argv=[]) == 0, 'Found errors'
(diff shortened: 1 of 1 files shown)


artivis:
@arnatious fyi

mikaelarguedas:
Is the problem also present on CentOS 8 ?

kyrofa:
Yeah +1 on the change in general, but I second @mikaelarguedas's question. This is an unversioned check, are we sure this is broken across all centos releases?

clalancette:
> Yeah +1 on the change in general, but I second @mikaelarguedas's question. This is an unversioned check, are we sure this is broken across all centos releases?

No, I'm not sure of that.  But my thinking here is that we are already doing static analysis of the mypy types on Ubuntu, macOS, and Windows.  There isn't really benefit to also running it on CentOS (regardless of the version), in my opinion.

mikaelarguedas:
> I'll run CI on this once I get agreement that this is the right way to go.

I'm no up-to-date regarding the CentOS story.
AFAICT from the REPs no current or upcoming ROS distro targets that platform. And one could assume that if a future ROS distro starts targeting CentOS it may target the one from 2019 and not from 2014, won't it ? 

Is the goal of this PR that CentOS users compiling from source can run tests locally ? or that the nightly job becomes green ?

If the former I'd argue that this is a reason for making the patch as narrow as necessary so that users can run all tests compatible with their setup. 

If the latter and following your point:

> But my thinking here is that we are already doing static analysis of the mypy types on Ubuntu, macOS, and Windows. There isn't really benefit to also running it on CentOS (regardless of the version), in my opinion.

:thinking: Would it be fair to say that all linter tests can be skipped on CentOS as they are ran on all the other platforms ?
If that is the case, an alternative is for the nightly jobs to pass de relevant flag for skipping those without having to add workaround code in various repos (that will then need to be maintained in multiple places).

clalancette:
> AFAICT from the REPs no current or upcoming ROS distro targets that platform. And one could assume that if a future ROS distro starts targeting CentOS it may target the one from 2019 and not from 2014, won't it ?

I would think that we would target CentOS 8 instead of 7 going forward, yes.  But as of today, the buildfarm is testing CentOS 7 nightly.

> Is the goal of this PR that CentOS users compiling from source can run tests locally ? or that the nightly job becomes green ?

A little bit of both, but mostly to make the nightly job green.

> Would it be fair to say that all linter tests can be skipped on CentOS as they are ran on all the other platforms ?
> If that is the case, an alternative is for the nightly jobs to pass de relevant flag for skipping those without having to add workaround code in various repos (that will then need to be maintained in multiple places).

That is a good point; we could do that.  If the problems were widespread, I'd probably consider doing that instead.  But in this case, this is the only test that is causing problems on CentOS, so a more targeted fix seems appropriate.

Overall, I'm hearing that we should narrow this to CentOS 7.  So I'll go ahead and do that for now.

clalancette:
239f197 makes it only skip the test on CentOS 7.

mikaelarguedas:
Thanks for the update. I opened https://github.com/ros2/sros2/pull/248 to avoid implementing a custom parsing logic for the os-release file, PTAL

clalancette:
#248 looks good to me, so closing this one out.

PR #257, opened 2021-04-09
Bump discovery delay when using RTI Connext-based RMWs

hidmic:
First dumb attempt to deal with `test_generate_policy` flakes when running against `rmw_connextdds` (e.g. https://ci.ros2.org/view/nightly/job/nightly_osx_repeated/lastCompletedBuild/testReport/sros2.test.sros2.commands.security.verbs/test_generate_policy/test_generate_policy/).

CI up to `sros2` against `rmw_connextdds` only:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14336/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9112/) <-- expected, no `rmw_connextdds` on `aarch64`
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12011/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14430/)

### DIFF
--- a/sros2/test/sros2/commands/security/verbs/utilities/sros2_cli_test_case.py
+++ b/sros2/test/sros2/commands/security/verbs/utilities/sros2_cli_test_case.py
@@ -63,7 +63,7 @@ class SROS2CLITestCase(unittest.TestCase):
         rmw_implementation,
         use_daemon
     ):
-        max_discovery_delay = 6.0 if 'connext' in rmw_implementation else 3.0  # seconds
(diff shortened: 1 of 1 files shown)


clalancette:
Still flaked on Linux :(.

hidmic:
Let's try another bump. It's not a real solution, but there's no way we can do an integration test for this command that's not sensitive to discovery latency.

hidmic:
I suspect this test is wrong. Waiting for the test itself to discover the fixture nodes does not imply the daemon has. Dropping in favor of #260.

PR #260, opened 2021-04-12
Wait for daemon to perform discovery unconditionally.

hidmic:
Alternative to #257. This patch introduces an unconditional delay for the daemon to discover fixture nodes.

CI up to `sros2`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14345/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9126/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12025/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14444/)

### DIFF
--- a/sros2/test/sros2/commands/security/verbs/test_generate_policy.py
+++ b/sros2/test/sros2/commands/security/verbs/test_generate_policy.py
@@ -23,14 +23,12 @@ import unittest
 from launch_ros.actions import Node
 
 import launch_testing
-from launch_testing.actions import ReadyToTest
 import launch_testing.asserts
 import launch_testing.markers
 import launch_testing.tools
 
 import pytest
 
-from rclpy.expand_topic_name import expand_topic_name
 from rclpy.utilities import get_available_rmw_implementations
 from sros2.policy import load_policy
 
@@ -87,8 +85,7 @@ def generate_test_description(rmw_implementation: str, use_daemon: bool):
     return generate_sros2_cli_test_description(
         fixture_actions=[
             pub_sub_node,
-            client_srv_node,
-            ReadyToTest()
+            client_srv_node
         ],
         rmw_implementation=rmw_implementation,
         use_daemon=use_daemon
@@ -106,11 +103,6 @@ class TestSROS2GeneratePolicyVerb(SROS2CLITestCase):
         pub_sub_node_namespace,
         pub_sub_node_enclave
     ):
-        assert self.wait_for(expected_topics=[
-            expand_topic_name('~/pub', pub_sub_node_name, pub_sub_node_namespace),
-            expand_topic_name('~/sub', pub_sub_node_name, pub_sub_node_namespace),
-        ])
-
         with tempfile.TemporaryDirectory() as tmpdir:
             test_policy = pathlib.Path(tmpdir).joinpath('test-policy.xml')
             with self.launch_sros2_command(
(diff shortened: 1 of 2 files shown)


sros2/test/sros2/commands/security/verbs/utilities/sros2_cli_test_case.py
```diff
@@ -37,13 +35,27 @@ def generate_sros2_cli_test_description(
...
-            additional_env=additional_env
-        )]
+        fixture_actions = [
+            ExecuteProcess(
+                cmd=['ros2', 'daemon', 'start'],
+                name='daemon-start',
+                on_exit=fixture_actions,
+                additional_env=additional_env
+            )
+        ]
```
clalancette, line 55: I think I'm getting lost in a maze of `fixture_actions`.  Isn't this assignment completely overwriting the previous `fixture_actions`, and thus there is no real `on_exit` handler happening here?  Relatedly, passing and changing a list to `generate_sros2_cli_test_description` seems to be an anti-pattern in Python, since it leads to surprising results.  Maybe it is worthwhile to clean this up and use some different variable names here?

clalancette (changes requested):
One place where I got bit lost in the code.  It may be worthwhile to clean that up.  Otherwise, I like the results!

sros2/test/sros2/commands/security/verbs/utilities/sros2_cli_test_case.py
hidmic: We can change variable names if it helps readability, for sure. But there's no mutation happening here. Variable assignment in Python does not mutate the variable value, it simply rebinds the variable name e.g.:

```python
>>> a = b = []
>>> a = [2, 3]
>>> b
[]
```

Mutation through assignment usually involves data descriptors or object protocols e.g.:

```python
>>> a = b = []
>>> a[:] = [2, 3]
>>> b
[2, 3]
```

clalancette: Yeah, you are totally right.  I was just getting a bit confused by the "chaining" of the variable name through, but it does work.

Personally I'd rename the variables to make it clear which ones were passed in vs. the ones we are generating in this method, but I won't block on it.

sros2/test/sros2/commands/security/verbs/test_generate_policy.py
```diff
@@ -106,11 +103,6 @@ def test_generate_policy_topics(
         pub_sub_node_namespace,
         pub_sub_node_enclave
     ):
-        assert self.wait_for(expected_topics=[
-            expand_topic_name('~/pub', pub_sub_node_name, pub_sub_node_namespace),
-            expand_topic_name('~/sub', pub_sub_node_name, pub_sub_node_namespace),
-        ])
```
clalancette, line 112: So I understand that this patch is remove `wait_for` on topics in favor of an unconditional 4 second wait.  However, this is additionally losing the check that ensures that those topics exist (which as I understand is a necessary, but not sufficient, precondition).  Should we check that the topics exist before moving on with the rest of the test?  Similar question below for the services.

clalancette (commented):
One more question.  Otherwise looks good to me.

sros2/test/sros2/commands/security/verbs/test_generate_policy.py
hidmic: > So I understand that this patch is remove wait_for on topics in favor of an unconditional 4 second wait.

That's right. 

> However, this is additionally losing the check that ensures that those topics exist (which as I understand is a necessary, but not sufficient, precondition)

These checks are useful when both test and `ros2` CLI use the daemon, as it speeds up the initial wait. When test and `ros2` CLI use their own node, it's useless (the fact that one discovered a topic or service does not imply the other has).

Thinking about this again, something's fishy. Flaky tests were often the ones using the daemon. Now I wonder if there's a connection with https://github.com/ros2/ros2cli/pull/620 I missed. I'll address that and see if it makes a difference here.

hidmic: Alright, I was indeed looking at the wrong place. The test implicitly assumes that topic or service discovery implies node discovery. That's not true. #260 fixes that by also waiting for the nodes, but only when a daemon is in use (it's pointless when not).

Issue #267, opened 2021-06-17
Specify Domain ID Through Command Line Argument


danthony06:
#### Feature description
Add the ability to specify the target domain through a command line argument, rather than environment variable. Currently, the domain ID for the configuration is set via reading an environment variable. This is clunky for automation and is harder to reproduce software between different developer's environments. Instead, it would be nice if this were set via a command line option to make it more explicit and easier to track.

#### Implementation considerations
It's currently set via an environment variable, but that is easy to forget about. A command line argument makes it explicit, but we can always use an environment variable as a fallback option.

ruffsl:
I think adding support a user provided config file would offer a more comprehensive approach to customization of the security certificate generation. This could extend beyond Domains ID, but also to certificate expiration dates, middleware presets, key storage formats, etc. A config file approach used by other tools in the ros ecosystem, like rviz, ros1_bridge, colcon, would be familiar to many ROS users, and could be a bit more sustainable than adding ever more niche or middleware specific command line argument options.

This is sort of akin to what I tried before with keymint:

* https://github.com/keymint/keymint_tools/tree/04a14e8da96e6e0c5fc16198b6b232b91d095f41
* https://github.com/ruffsl/keymint_ws/tree/cc9306117b12a61989466c2e728304ccd54dc65c/profile/package.defaults

PR #268, opened 2021-06-17
Add domain id argument

danthony06:
Addresses https://github.com/ros2/sros2/issues/267. I haven't worked with SROS2 before, so apologies in advance for anything I have messed up.

Distro A, OPSEC #4584

### DIFF
--- a/sros2/sros2/_utilities.py
+++ b/sros2/sros2/_utilities.py
@@ -39,7 +39,7 @@ def create_symlink(*, src: pathlib.Path, dst: pathlib.Path):
     os.symlink(src, dst)
 
 
-def domain_id() -> int:
+def environment_domain_id() -> int:
     return os.getenv(_DOMAIN_ID_ENV, 0)
 
 
--- a/sros2/sros2/_utilities.py
+++ b/sros2/sros2/_utilities.py
@@ -39,8 +39,11 @@ def create_symlink(*, src: pathlib.Path, dst: pathlib.Path):
     os.symlink(src, dst)
 
 
-def environment_domain_id() -> int:
-    return os.getenv(_DOMAIN_ID_ENV, 0)
+def get_domain_id(domain_id: int = None) -> int:
+    validated_domain = domain_id
+    if (domain_id is None) or (domain_id < 0):
+        validated_domain = os.getenv(_DOMAIN_ID_ENV, 0)
+    return  validated_domain 
 
 
 def get_keystore_path_from_env() -> pathlib.Path:
--- a/sros2/sros2/_utilities.py
+++ b/sros2/sros2/_utilities.py
@@ -39,8 +39,8 @@ def create_symlink(*, src: pathlib.Path, dst: pathlib.Path):
     os.symlink(src, dst)
 
 
-def domain_id() -> str:
-    return os.getenv(_DOMAIN_ID_ENV, '0')
+def domain_id() -> int:
+    return os.getenv(_DOMAIN_ID_ENV, 0)
 
 
 def get_keystore_path_from_env() -> pathlib.Path:
--- a/sros2/sros2/api/_artifact_generation.py
+++ b/sros2/sros2/api/_artifact_generation.py
@@ -25,7 +25,7 @@ def generate_artifacts(
         keystore_path: pathlib.Path = None,
         identity_names: List[str] = [],
         policy_files: List[pathlib.Path] = [],
-        domain_id: int = 0) -> None:
+        domain_id: int = None) -> None:
     if keystore_path is None:
         keystore_path = _utilities.get_keystore_path_from_env()
         if keystore_path is None:
--- a/sros2/sros2/api/_artifact_generation.py
+++ b/sros2/sros2/api/_artifact_generation.py
@@ -25,7 +25,7 @@ def generate_artifacts(
         keystore_path: pathlib.Path = None,
         identity_names: List[str] = [],
         policy_files: List[pathlib.Path] = [],
-        domain_id: int = None) -> None:
+        domain_id: int = 0) -> None:
     if keystore_path is None:
         keystore_path = _utilities.get_keystore_path_from_env()
         if keystore_path is None:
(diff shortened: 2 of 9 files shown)


sros2/sros2/api/_artifact_generation.py
```diff
@@ -24,18 +24,19 @@
 def generate_artifacts(
         keystore_path: pathlib.Path = None,
         identity_names: List[str] = [],
-        policy_files: List[pathlib.Path] = []) -> None:
+        policy_files: List[pathlib.Path] = [],
+        domain_id: int = 0) -> None:
```
mikaelarguedas, line 28: By setting the default value to 0 here, doesn't this bypass the fallback to the environment variable ? It should be None here too isn't it?


Adding a corresponding test to the test suite would be valuable to validate the behavior

mikaelarguedas:
Thanks for the PR! I'll try to find time to go over it this weekend.

Ideally the argument would be optional in the command line and all functions, otherwise it will be breaking API making that PR not suitable for galactic and would have to be held until the next ROS release.

sros2/sros2/api/_artifact_generation.py
danthony06: Good catch, I'll get this fixed and add test cases.

danthony06:
@mikaelarguedas I added a fallback so that if the command line argument is not provided, it should fall back and check the environment variable and maintain the same behavior. @ruffsl does not seem as excited by this, so do you think I should keep working on this?

sros2/sros2/_utilities.py
```diff
@@ -39,8 +39,11 @@ def create_symlink(*, src: pathlib.Path, dst: pathlib.Path):
     os.symlink(src, dst)
 
 
-def domain_id() -> str:
-    return os.getenv(_DOMAIN_ID_ENV, '0')
+def get_domain_id(domain_id: int = None) -> int:
+    validated_domain = domain_id
+    if (domain_id is None) or (domain_id < 0):
```
mikaelarguedas, line 44: if someone explicitly passed a negative domain ID, we should likely raise a ValueError to be explicit that this is not a valid value rather than silently replacing it with 0

mikaelarguedas, line 44:
```suggestion
    if domain_id is None:
```

sros2/sros2/keystore/_enclave.py
```diff
@@ -32,12 +32,15 @@
...
 
-def create_enclave(keystore_path: pathlib.Path, identity: str) -> None:
+def create_enclave(keystore_path: pathlib.Path, identity: str, domain_id: int) -> None:
     if not _keystore.is_valid_keystore(keystore_path):
         raise sros2.errors.InvalidKeystoreError(keystore_path)
 
     if not _is_enclave_name_valid(identity):
         raise sros2.errors.InvalidEnclaveNameError(identity)
+    
+    if domain_id <= 0:
```
mikaelarguedas, line 42: same here

sros2/sros2/keystore/_permission.py
```diff
@@ -84,7 +90,7 @@ def create_permission_file(path: pathlib.Path, domain_id, policy_element) -> Non
 
     domain_id_elements = permissions_xml.findall('permissions/grant/*/domains/id')
     for domain_id_element in domain_id_elements:
-        domain_id_element.text = domain_id
+        domain_id_element.text = str(domain_id)
```
mikaelarguedas, line 93: can you also specify the domain_id type in the function signature ?

sros2/sros2/verb/create_enclave.py
```diff
@@ -33,10 +33,12 @@ def add_arguments(self, parser, cli_name) -> None:
         arg = parser.add_argument('ROOT', type=pathlib.Path, help='root path of keystore')
         arg.completer = DirectoriesCompleter()
         parser.add_argument('NAME', help='enclave name')
+        parser.add_argument('--domain_id', '-d', dest='domain_id', type=int,
+            help='domain id for rules', default=-1, required=False, nargs=1)
```
mikaelarguedas, line 37: Why set an invalid default value here ?

sros2/sros2/verb/create_keystore.py
```diff
@@ -24,18 +24,28 @@ def DirectoriesCompleter():
...
 import sros2.keystore
 from sros2.verb import VerbExtension
+from sros2 import _utilities
 
 
 class CreateKeystoreVerb(VerbExtension):
     """Create keystore."""
 
     def add_arguments(self, parser, cli_name) -> None:
+        parser.add_argument('--domain', dest='DOMAIN_ID', type=self._validate_domain_id, default=self._validate_domain_id(),
```
mikaelarguedas, line 34: why call the argument `domain` here and `domain_id` in the creat_enclave verb ?

sros2/sros2/verb/create_keystore.py
```diff
@@ -24,18 +24,28 @@ def DirectoriesCompleter():
...
+            required=False, help='domain ID of keystore')
+
         arg = parser.add_argument('ROOT', type=pathlib.Path, help='root path of keystore')
         arg.completer = DirectoriesCompleter()
+    
+    def _validate_domain_id(self, domain_id: int = None) -> int:
+        if domain_id is None:
+            return _utilities.get_domain_id()
+        else:
+            return _utilities.get_domain_id(int(domain_id))
```
mikaelarguedas, line 44: why call get_domain_id here ?

mikaelarguedas (commented):
This PR would need to:
pass the tests including DCO signature
Update existing test cases to test both provided arguments and default behavior
Add new test cases where relevant to cover the new feature

> @ruffsl does not seem as excited by this, so do you think I should keep working on this?

@ruffsl do you think such config file will be implemented in the near future ? Would you consider it a blocker for this incremental improvement ?

ruffsl:
> @ruffsl do you think such config file will be implemented in the near future ? Would you consider it a blocker for this incremental improvement ?

We could go through with this incremental addition to the CLI, but I'd just like to ensure we are comfortable reverting this once a proper user config file is supported. Adding transport specific arguments to the CLI just doesn't seem very scalable.

I could also propose a config format during the next Security WG meeting, but in the meantime I think we should step back and do a compressive audit of all the finer parameters and settings we'd like to expose or default for end users, e.g. Secure DDS governance options, CA hierarchy, etc.

PR #272, opened 2021-07-08
Add in some documentation about how to use CRLs on Linux.

clalancette:


clalancette:
Friendly ping on this; I'd like to get some documentation about this feature landed.  Thanks!

clalancette:
Going ahead and merging with one review; since it is a documentation-only change, it should be safe enough.  Also not running CI and relying on the PR checks for the same reason.

PR #276, opened 2022-01-13
Fast-forward galactic branch

jacobperron:
The recent changes on `master` easily apply to Galactic.


SROS2_Linux.md
```diff
@@ -204,3 +204,62 @@ For example, the following attempt for the `listener` node to subscribe to a top
...
+You'll notice that this is operating just like before.
+
+In another terminal (after preparing the terminal as previously described), we will do the same thing with the `listener` program.
+
+```bash
+ros2 run demo_nodes_py listener --ros-args --enclave /talker_listener/listener
+```
+
+Here you'll notice that the listener is not getting any data.
+That's because the talker is explicitly rejecting the listener revoked certificate of the listener, so no communication is possible.
```
audrow, line 265: nit: no new line at the end of the file.
```suggestion
That's because the talker is explicitly rejecting the listener revoked certificate of the listener, so no communication is possible.

```

jacobperron: Since my intention is to fast-forward commits from `master` (keeping identical commit hashes), I'd rather not modify contents in this PR.

audrow: Sounds good

PR #290, opened 2024-01-31
Use modern PKCS7 to sign the certificate bytes.

clalancette:
Using the SSLBinding leads to a warning in newer versions of pycryptography.  Luckily, there is a supported API called pkcs7 that allows us to do the same thing.  Even better, this API is supported since pycryptography 3.2, so this should work on both Ubuntu 22.04 and Ubuntu 24.04 without warnings.

### DIFF
--- a/sros2/sros2/_utilities.py
+++ b/sros2/sros2/_utilities.py
@@ -19,10 +19,10 @@ import pathlib
 
 from cryptography import x509
 from cryptography.hazmat.backends import default_backend as cryptography_backend
-from cryptography.hazmat.bindings.openssl.binding import Binding as SSLBinding
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import ec
+from cryptography.hazmat.primitives.serialization import pkcs7
 
 import sros2.errors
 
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20194/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14614/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20905/)

clalancette:
OK, we need to keep the fallback path for Windows.  I've done that in 389dc4c, here is new CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20195/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14615/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20906/)

marcoag:
I gave it a try and works for me. Thanks for the much needed change!

marcoag (approved):

mikaelarguedas:
@clalancette thanks for this!

A couple follow-up questions
- can the issue https://github.com/ros2/sros2/issues/239 be closed now that this is merged ?
- should this be backported to iron and humble ?
- For what active ROS distros the SSL bindings are still useful ? (I suppose None once https://github.com/ros2/ci/pull/778 is in?)

clalancette:
> can the issue https://github.com/ros2/sros2/issues/239 be closed now that this is merged ?

Yep!  Done now.

> * should this be backported to iron and humble ?

I don't think we particularly need it there.  We are still using the older version of pycryptography there, which has the old API.

> * For what active ROS distros the SSL bindings are still useful ? (I suppose None once https://github.com/ros2/ci/pull/778 is in?)

No, we still need it on Windows, even on Jazzy and Rolling.  We are still using the older version of pycryptography there.

mikaelarguedas:
> I don't think we particularly need it there. We are still using the older version of pycryptography there, which has the old API.

Ok sorry for catching the train on route so long after. My understanding from reading the issues was that the reason for this PR was to support modern versions of cryptography as on Windows the version was too recent and caused breakage on Iron https://github.com/ros2/ros2_documentation/issues/4069.
So if this API is not made avialable on Iron, that means this issue is still open and users are still force to downgrade cryptography on their system https://github.com/ros2/ros2_documentation/blob/fa244408830942eb45b32a31ae6b11d79320571a/source/Releases/Release-Iron-Irwini.rst?plain=1#L849 is that correct ?

> No, we still need it on Windows, even on Jazzy and Rolling. We are still using the older version of pycryptography there.

Is there something specific holding us back here ?

It looks like cryptography is compatible with python 3.8,
the system one on Ubuntu 24.04 is `41.0.7-4build3`
on ubuntu 22.04 `3.4.8-1ubuntu2.2`
on windows it's installed from pip (so way above the faulty 3.1.0 version)

clalancette:
> Is there something specific holding us back here ?

Yes, Windows Debug.

We have an ancient version of pycryptography that we have built by hand, and have not updated in a long time; you can see it get installed here: https://github.com/ros2/ci/blob/3435d04acf17872b07861e81265bc34d7ef40c72/ros2_batch_job/__main__.py#L493

The correct solution here would be to update that debug wheel to a modern pycryptography version, but a) it needs to build against Python 3.8 for the other reasons mentioned, and b) when I poked at it a couple of months ago, I couldn't figure out how to make it build at all.  If you have time to look into that, it would be appreciated.

PR #299, opened 2024-05-06
[CI] rolling is now the default branch, update CI and README accordingly

mikaelarguedas:
Also bumps runner to ubuntu-latest but doesnt matter as everything runs in containers

This is an attempt to revive CI that has been off for a long time

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
         apt-get -qq update
         apt-get -qq upgrade -y
         apt-get -qq install -y curl libasio-dev libssl-dev libtinyxml2-dev python3-colcon-coveragepy-result
-    - uses: ros-tooling/action-ros-ci@v0.1
+    - uses: ros-tooling/action-ros-ci@v0.3
       with:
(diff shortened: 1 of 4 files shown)


fujitatomoya (approved):
@mikaelarguedas thanks for the PR.

as you mentioned, `runs-on` does not really matter, since container is used anyway. (just fyi, https://github.com/actions/runner-images?tab=readme-ov-file#available-images, ubuntu-latest is still ubuntu 22.04.)

and nightly build looks good to me, but we do not have any CI for `iron` and `humble`? i did not really know the history for sros2 CI...

mikaelarguedas:
> and nightly build looks good to me, but we do not have any CI for iron and humble? i did not really know the history for sros2 CI...

my feeling is that this workflow has 2 main goals ->
- test_latest: generate coverage on pull requests and merges
- test_nightly: catch any regression overnight

for released distros I believe we can keep relying on the PR jobs of the ROS buildfarm :+1:

mikaelarguedas:
I'll get this in as-is and work on gettign latest back to working in a separate PR

PR #303, opened 2024-05-08
add back codecov.yml config

mikaelarguedas:
Roll back https://github.com/ros2/sros2/pull/185
Using new `codecov_yml_path` argument: https://github.com/codecov/codecov-action?tab=readme-ov-file#arguments

This should fix coverage visualization on codecov.io

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,6 +51,7 @@ jobs:
         flags: unittests
         name: sros2-coverage
         fail_ci_if_error: true
+        codecov_yml_path: codecov.yml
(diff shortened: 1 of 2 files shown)


ahcorde (approved):

fujitatomoya (approved):
lgtm

mikaelarguedas:
@ahcorde @fujitatomoya thanks for the reviews!

The only change I made since your review was that I adjusted the path rewriting in the codecov.yml file

This seems to now work as expected, we'll be sure only once we get coverage on rolling with that change and PRs coming in.

PR #304, opened 2024-05-08
move coverage to test_nightly

mikaelarguedas:
as coverage doesnt depend on dependencies and test_latest takes forever


mikaelarguedas:
for some reason I have the same odd behavior locally:
- if installed from packages or source: coverage sros2 coverage 78%
- if using the nightly archive and building 

While test_generate policy passes it doesnt appear in coverage results.

Stalling this for now

Issue #306, opened 2024-05-08
Sometimes `ROS_SECURITY_ENCLAVE_OVERRIDE` does not effectively work as introduced


fujitatomoya:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 24.04
- Installation type:
  - source build
- Version or commit hash:
  - https://github.com/ros2/ros2/commit/4d07f58591920794676261a14c2ba9d0a7bf2746
- DDS implementation:
  - Fast-DDS likely, Cyclonedds sometimes
- Client library (if applicable):
  - rclpy, ros2cli

#### Steps to reproduce issue

https://github.com/ros2/sros2/blob/rolling/SROS2_Linux.md#run-the-demo-1 introduces `ROS_SECURITY_ENCLAVE_OVERRIDE` to be used with `ros2cli` such as `ros2 node list` and `ros2 topic list`.
But sometimes it does not print the available nodes or topics as explained.

- start `talker` and `listener`, make sure they are working as expected.

```shell
ros2 run demo_nodes_cpp talker --ros-args --enclave /talker_listener/talker
ros2 run demo_nodes_py listener --ros-args --enclave /talker_listener/listener
```

- call `ros2cli` as followings.

```shell
root@tomoyafujita:~/ros2_ws/colcon_ws# ros2 node list
root@tomoyafujita:~/ros2_ws/colcon_ws# ros2 topic list
/parameter_events
/rosout
```

- even with `--no-daemon`, it sometimes does not work.

```shell
root@51cdd59e1f3e:~# ros2 node list --no-daemon
[INFO] [1715030178.386225737] [rcl]: Found security directory: /root/sros2_demo/demo_keystore/enclaves/talker_listener/listener
root@51cdd59e1f3e:~# ros2 topic list --no-daemon
[INFO] [1715030183.341342733] [rcl]: Found security directory: /root/sros2_demo/demo_keystore/enclaves/talker_listener/listener
/parameter_events
/rosout
```

#### Expected behavior

node list and topic list should print available nodes (talker and listener) and topic (/chatter).

#### Actual behavior

described above.

#### Additional information

originally found on https://github.com/ros2/sros2/pull/295

fujitatomoya:
So this is all about the `ros2 daemon`.

```shell
root@tomoyafujita:~/ros2_ws/colcon_ws# env | grep ROS
ROS_VERSION=2
ROS_SECURITY_ENCLAVE_OVERRIDE=/talker_listener/listener
[...]

PR #307, opened 2024-05-08
add `--no-daemon` and `--spin-time` explanation in readme.

fujitatomoya:
address part of https://github.com/ros2/sros2/issues/306

### DIFF
--- a/SROS2_Linux.md
+++ b/SROS2_Linux.md
@@ -126,13 +126,13 @@ Then use the CLI with `--no-daemon` and `--spin-time`:
 > Avoid using `ros2 daemon` since it may not have security enclaves, and enough time duration should be given for the discovery in secured network.
 
 ```bash
-ros2 node list --no-daemon --spin-time 3
+ros2 node list --no-daemon --spin-time 4
 ```
 ```
 /talker
 ```
 ```bash
-ros2 topic list --no-daemon --spin-time 3
+ros2 topic list --no-daemon --spin-time 4
 ```
 ```
 /chatter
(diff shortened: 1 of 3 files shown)


SROS2_Linux.md
```diff
@@ -120,24 +120,27 @@ To be able to use the ros2 CLI tools to interact with your secured system, you n
...
 
-Then use the CLI as usual:
+Then use the CLI with `--no-daemon` and `--spin-time`:
+
+> [!NOTE]
+> Avoid using `ros2 daemon` since it may not have security enclaves, and enough time duration should be given for the discovery in secured network.
 
 ```bash
-ros2 node list
+ros2 node list --no-daemon --spin-time 3
```
fujitatomoya: 3 seconds come from my amd64 environment, it should give it enough time window for the security handshake and discovery. (actually, fast-dds security takes much longer time compared to cyclonedds, that is why we did not see this issue with cyclonedds. but possibly this is common issue.)

fujitatomoya (commented):
@mikaelarguedas could you review this?

fujitatomoya:
i think we want to backport this to jazzy at least.

SROS2_Linux.md
mikaelarguedas:
```suggestion
ros2 node list --no-daemon --spin-time 4
```

Unclear how the value was picked at the time, but in the integration tests the value used is 4 so maybe we should use the samme here ?
https://github.com/ros2/sros2/blob/07608ab784bac6b295e7c031692f9bc6598adf76/sros2/test/sros2/commands/security/verbs/utilities/sros2_cli_test_case.py#L32
https://github.com/ros2/sros2/blob/07608ab784bac6b295e7c031692f9bc6598adf76/sros2/test/sros2/commands/security/verbs/utilities/sros2_cli_test_case.py#L72-L74

fujitatomoya: sounds good.

mikaelarguedas (approved):
:+1: lgtm and agree to backport to jazzy as this bit of the tutorial alreay made it to the Jazzy branch: https://github.com/ros2/sros2/pull/296

mikaelarguedas:
@Mergifyio backport jazzy

PR #312, opened 2024-05-10
[CI] remove obsolete / broken workflow

mikaelarguedas:
As of May 7th 2024 the github action workflow "SROS2 CI" has been reenabled.

But this workflow is meant to test only nightly and latest source builds so that applies only to the rolling branch and will consistently fail on other distro branches.

This PR removes the outdated workflow file on Jazzy and is meant to be backported to all active distro branches


mikaelarguedas:
@Mergifyio backport iron humble

PR #317, opened 2024-05-22
Fix sros2 tests on Windows Debug.

### LINKED ISSUE #316: :woman_farmer: Windows debug `Windows fatal exception: access violation` build regression
## Bug report

**Required Info:**

- Operating System: 
  - Windows
- Installation type: 
  - from source
- Version or commit hash:
  - https://github.com/ros2/sros2/commit/fb22661623426fadcc60804e10a964664330d98e

#### Steps to reproduce issue
Run a windows debug build on ci.ros2.org.

#### Expected behavior
For the build to pass unstable

#### Actual behavior
Build fails

#### Additional information

Reference build: https://ci.ros2.org/view/nightly/job/nightly_win_deb/3095/
Failing since May 16th. Due to the timing and content of the merged PR we think #311 introduced this build regression.

**Error**: 
```
04:05:34 test\sros2\commands\security\verbs\test_create_permission.py .           [ 65%]

04:05:34 Windows fatal exception: access violation
04:05:34 
04:05:34 Current thread 0x00002c94 (most recent call first):
04:05:34   File "C:\ci\ws\src\ros2\sros2\sros2\sros2\verb\generate_artifacts.py", line 44 in main
```

clalancette:
From what I can tell, it looks like lxml has a bug where it doesn't properly track references to objects via the find() method.  This manifests on Windows debug as a crash *after* we have stopped using the object, but I believe that by that point the underlying memory has already been freed.  Windows Debug in particular is sensitive to this.

Fix it by doing a deepcopy of the object returned from the find().  This code isn't performance sensitive, so it shouldn't be a big deal to do it here, and it fixes the bug in my testing.

### DIFF
--- a/sros2/sros2/api/_policy.py
+++ b/sros2/sros2/api/_policy.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from copy import deepcopy
+
 from lxml import etree
 
 from sros2.policy import load_policy
@@ -28,7 +30,7 @@ def get_policy_from_tree(name, policy_tree):
     if enclave_element is None:
         raise RuntimeError(f'unable to find enclave "{name}"')
     enclaves_element = etree.Element('enclaves')
-    enclaves_element.append(enclave_element)
+    enclaves_element.append(deepcopy(enclave_element))
     policy_element = etree.Element('policy')
     policy_element.append(enclaves_element)
     return policy_element


mikaelarguedas:
Thanks 🙏 

Ideally I'd like to keep separate the behavioral fix and API/signature change to streamline fix back porting to older distros.
Regarding the mutable defaults 👍 (hence the frecebtly added file), however what would you think about making all ARGS keyed-args and keeping the order as is ? 

Finally is it possible to trigger a windows debug job once the PR splitted to make sure the CI is happy about it ?

clalancette:
> however what would you think about making all ARGS keyed-args and keeping the order as is ?

Actually, to be perfectly frank we could just leave the order as-is and remove all of the default arguments.  This function is always called with all of its arguments.  So I think I'll open a separate PR to do that.

clalancette:
Here is just CI on Windows Debug.  Assuming this is successful, I'll run full CI on it:

* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/21856/)

clalancette:
And see #318 for the rest of the fixes.

fujitatomoya (approved):
lgtm with green CI, thanks for the fix.

clalancette:
And here is full CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21133/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15433/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21857/)

mikaelarguedas:
Cheers!

mikaelarguedas (approved):

mikaelarguedas:
@Mergifyio backport jazzy

PR #318, opened 2024-05-22
[FIX] remove dangerous mutable default arguments in generate_artifacts

clalancette:
You cannot safely use = [] as a default argument in Python; subsequent calls will reuse that object.  It probably doesn't matter in this case, as these are all "read-only" arguments, but it is better not to do it at all.  Instead, we recognize that this function is always called with all of its arguments, so there is no need for default arguments at all.

While in here, I noticed that, generate_artifacts was returning a boolean on an error path, which is not allowed by its signature.  So fix that as well.

### DIFF
--- a/sros2/sros2/api/_artifact_generation.py
+++ b/sros2/sros2/api/_artifact_generation.py
@@ -21,16 +21,15 @@ from sros2.policy import load_policy
 from . import _policy
 
 
-# FIXME move away from mutable default (linter should complain about it)
 def generate_artifacts(
-    keystore_path: Optional[pathlib.Path] = None,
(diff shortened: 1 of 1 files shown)


fujitatomoya (approved):
lgtm!

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21135/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/15436/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21858/)

mikaelarguedas (approved):

PR #327, opened 2024-12-18
Use multiple CAs instead of a single self-signed root CA

Santti4go:
Ticket https://github.com/ros2/sros2/issues/328

I added a new (optional) flag for the `create_keystore` command: `--split-CA`
This flag changes the Certificate Authorities structure creating two new CAs (Permissions CA and Identity CA) instead of using the same self-signed root CA with symlinks.
This is an optional flag and does not change default behavior.

More details in the ticket https://github.com/ros2/sros2/issues/328

### DIFF
--- a/sros2/sros2/_utilities.py
+++ b/sros2/sros2/_utilities.py
@@ -28,6 +28,33 @@ import sros2.errors
 _DOMAIN_ID_ENV = 'ROS_DOMAIN_ID'
 _KEYSTORE_DIR_ENV = 'ROS_SECURITY_KEYSTORE'
 
+def create_signed_cert(
+        keystore_ca_cert_path: pathlib.Path,
+        keystore_ca_key_path: pathlib.Path,
+        identity: str,
+        cert_path: pathlib.Path,
+        key_path: pathlib.Path,
+        **kwargs):
+    # Load the CA cert and key from disk
+    ca_cert = load_cert(keystore_ca_cert_path)
+
+    with open(keystore_ca_key_path, 'rb') as f:
+        ca_key = serialization.load_pem_private_key(f.read(), None, cryptography_backend())
+
+    ca_pub_key = ca_cert.public_key()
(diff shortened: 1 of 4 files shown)


fujitatomoya (commented):
i think these kind of new arguments or feature need to be supported in rolling 1st, and then we can consider backport for already released distros including humble.

Santti4go:
> i think these kind of new arguments or feature need to be supported in rolling 1st, and then we can consider backport for already released distros including humble.

Sure, I targeted Humble because it's the branch I'm working on.
If we agree this is a feature we want to add I'll happily target Rolling instead. 
WDYT? Do I have green light?

fujitatomoya:
to be honest, i am not sure. having this optional mode seems to be okay but i do not really maintain this repository. @mikaelarguedas could you take a look or ping someone else here?

Santti4go:
Friendly ping @mikaelarguedas 

I've made a few more changes on my end. If we agree this is something we want to support I would happily push them here and target Rolling and then backport it. Let me know what you think.

Santti4go:
So, I rebased into rolling as you @fujitatomoya asked. Then I would like to backport to Humble as well.
While I was here I also added a `path_length` attribute for the CA certificates and some other minor changes like adding full chain for Permissions and Identity CAs.

CC @mikaelarguedas

fujitatomoya:
@Santti4go i guess having the option for this should be okay, but i would like to recommend you to bring this topic to https://discourse.ros.org/t/security-working-group-meeting-february-2025/41957 security working group to discuss.

audrow:
What were the outcomes of the security working group?

Santti4go:
> What were the outcomes of the security working group?

Last week -Feb 11th- I raised the topic with Flor Cabral during the meeting. Since it was just the two of us, we decided to continue any discussion that might come up here.

Do you have any thoughts on the PR content? @audrow

PR #330, opened 2025-01-24
Create new rmw_security_common

ahcorde:
the new `rmw_zenoh_cpp` has an open PR to include certificates using the `--enclave` argument. This requires to use `rmw_dds_common::get_security_files` which is not a dds unique utility. 

The idea with new package is to include here all the security common things.

### DIFF
--- a/rmw_security_common/CMakeLists.txt
+++ b/rmw_security_common/CMakeLists.txt
@@ -1,0 +1,64 @@
+cmake_minimum_required(VERSION 3.5)
+
+project(rmw_security_common)
+
+# Default to C++17
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 17)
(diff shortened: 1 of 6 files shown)


clalancette:
I know I suggested this, but I am slightly concerned here about the fact that this package is very low-level (e.g. used by RMW implementations), while the rest of the packages here are very high-level (e.g. Python and CMake packages).  I discussed this a bit with @mjcarroll , and we thought it might make more sense to actually move this into https://github.com/ros2/rmw .  However, as part of that it might actually also make sense to reimplement it as a pure-C header, rather than C++ (so it can be wrapped).

I'm going to put this on the agenda for the ROS 2 weekly meeting next week so we can discuss it a bit more.

fujitatomoya (commented):
> we thought it might make more sense to actually move this into https://github.com/ros2/rmw

i agree with this 👍

ahcorde:
This PR is replaced by https://github.com/ros2/rmw/pull/388

PR #335, opened 2025-03-09
[DEV-OPS] add dependabot file to open PRs to bump github actions versions

mikaelarguedas:


fujitatomoya (approved):
this is nice to have, especially for `security` package. lgtm 👍

.github/dependabot.yml
```diff
@@ -0,0 +1,8 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
christophebedard, line 6: In my experience, using a daily interval like this quickly becomes annoying, unless you really don't have that many dependencies. It seems like that might be the case, though, since there's only 1 pretty small file (`.github/workflows/test.yml`).

PR #394, opened 2026-07-09
Type hints

ahcorde:
## Description

Type hints

### Did you use Generative AI?
Claude Opus 4.7


fujitatomoya (commented):
does this actually pass mypy?

AFAIS, in lxml those are factory functions, not classes? the actual classes are `etree._Element` and `etree._ElementTree`. adding these type checks would look ugly a bit, but i believe those types should be used here?

